PR #3, opened 2017-07-13
Make unpack_nested_data a lot more flexible with supported column types

austin3dickey:
Now this can handle columns that are arrays of simple objects like strings, rather than just arrays of maps.


austin3dickey:
Hold on, I need to add two quick tests.

PR #6, opened 2017-07-13
Fix doc and DESCRIPTION info

staftermath:
Made the following changes to pass `R CMD check`
1. Edited DESCRIPTION and .md link
2. Allow 1 captured warning msg from `R CMD check --as-cran --no-manual` in case library MASS is not found
3. `devtools::check` still have NOTE on title, specifically `data.table` to `Data.table`, deliberately ignored this one.
4. Remove unnecessary `lubridate` in test


r-pkg/uptasticsearch/tests/testthat/test-repo_characteristics.R
```diff
@@ -39,7 +39,7 @@ test_that("R CMD check should not return any errors", {
 })
 
 test_that("R CMD check should not return any warnings", {
-    expect_true(length(x[["warnings"]]) == 0)
+    expect_true(length(x[["warnings"]]) == 1)
```
jameslamb: Can you please add a comment showing the 1 warning we are tolerating and explaining the conditions under which it can arise (i.e. on a system without pdflatex)?

Also would like @austin3dickey 's opinion on this

staftermath: Good call! Comment edited (item 2).

r-pkg/uptasticsearch/tests/testthat/test-repo_characteristics.R
```diff
@@ -39,7 +39,7 @@ test_that("R CMD check should not return any errors", {
 })
 
 test_that("R CMD check should not return any warnings", {
-    expect_true(length(x[["warnings"]]) == 0)
+    expect_true(length(x[["warnings"]]) <= 1)
```
austin3dickey: I think this is unnecessary, and it should be `== 0`. On a system without `pdflatex`, it flags a warning in its verbosity, but does not store the WARNING in the final result of R CMD check (i.e. `x[["warnings"]]` should be an empty character). 

If that's not true on someone else's system, please let me know. And if that's the case, we should make this test something like `expect_true(length(x[["warnings"]]) == 0 || x[["warnings"]] == "warning you don't have latex ya noob")`.

jameslamb: agree completely, thanks Austin!

staftermath: the test itself runs fine. I was checking using option `--as-cran --no-manual` and it pops a warning. Changed it back now.

austin3dickey:
I think you should change the test to `expect_true(length(x[["warnings"]]) == 0 || x[["warnings"]] == "checking Rd cross-references ... WARNING\nError in find.package(package, lib.loc) : \n  there is no package called ‘MASS’\nCalls: <Anonymous> -> lapply -> FUN -> find.package\nExecution halted")`.... JUST in case.

PR #8, opened 2017-07-14
Added pkgdown docs

austin3dickey:
Added a site we can put somewhere. Look at it at `r-pkg/uptasticsearch/docs/index.html`.

I also added a NEWS.md and README.md in the package's directory because they make rendering the site a lot easier. The new README is a copy of the top-level README with references to Python removed.


terrytangyuan:
Looks good to me! BTW, I think we should probably remove the redundant `uptasticsearch` folder so we will only need uptasticsearch/r-pkg/ that contains the actual R package.

r-pkg/uptasticsearch/NEWS.md
```diff
@@ -0,0 +1,19 @@
+# uptasticsearch 0.1.0
+
+## First release!
```
terrytangyuan: Maybe remove this line and the description for the package below. Put the other sub sections inside a section named like `Features`.

austin3dickey: Did I make this change correctly?

terrytangyuan: Yeah thanks!

PR #19, opened 2017-07-18
Changes following first CRAN submission

jameslamb:


README.md
```diff
@@ -23,10 +23,10 @@ The core functionality of this package is the `es_search` function. This returns
 
 ### R <a name="rinstallation"></a>
 
-The package is not yet on CRAN, but we'll be submitting soon. For now, you can install using `devtools`.
+This package can be installed directly from CRAN.
```
terrytangyuan: We should also keep the instruction for installing dev version for adventurous users

jameslamb: Good idea! Will fix

README.md
```diff
@@ -1,5 +1,7 @@
 # uptasticsearch
 
+[![CRAN\_Status\_Badge](http://www.r-pkg.org/badges/version-last-release/uptasticsearch)] [![CRAN\_Download\_Badge](http://cranlogs.r-pkg.org/badges/grand-total/uptasticsearch)]
```
austin3dickey: Can you add the badges in the `r-pkg/README` as well and regen `/docs/` again?

austin3dickey: Hopefully pkgdown will add the badges to its sidebar; that's the intended feature

PR #22, opened 2017-07-20
clarified documentation for max_hits default parameter and then chang…

mfrasco:
…ed default parameter from NULL to Inf


r-pkg/R/elasticsearch_parsers.R
```diff
@@ -680,7 +682,9 @@ es_search <- function(es_host
 #               time you expect to pass between requests. See the 
 #               \href{https://www.elastic.co/guide/en/Elasticsearch/guide/current/scroll.html}{Elasticsearch scroll/pagination docs}
 #               for more information.
-# [param] max_hits Integer. If specified, \code{.fetch_all} will stop pulling data as soon as it passes this threshold.
+# [param] max_hits Integer. If specified, \code{es_search} will stop pulling data as soon
+#'                 as it has pulled this many hits. Default is \code{Inf}, meaning that
```
austin3dickey: Can you remove the roxygen apostrophes after the hashtags here since this isn't officially Roxygen?

mfrasco: Done. Thanks for catching this.

PR #23, opened 2017-07-21
Feature/retrieve mapping

mfrasco:
Let me know what you guys think about the initial draft of this function. I think that the `unlist` is pretty clever. My first version had the ugliest lapply ever. I tried this out on elasticsearch 5.5 and it works. There are a few things that could be tweaked. I'll add unit tests once you give me the go-ahead that the function has life.


jameslamb:
I think this is a cool utility. Some things we need to work out:

1. This is catching "store", "analyzer", and "format" entries in the mapping as if those were fields. It should not be. If I have a mapping for "field1.field2" with an entry `"store": True`, that doesn't mean there is a field called "field1.field2.store".
2. Drop the `es_types` argument. This function will run pretty fast. People can do their own postprocessing if they want to drop types. It's not intuitive to me how multiple es_types would work with multiple indices anyway.
3. Don't worry about the datatype. Right now you're getting a lot of bad answers for data type (like `TRUE` and `not_analyzed`). I think we should restrict this function to listing the names of indexed fields in each index.
4. Please clarify in the function docs that this will return only the names of **indexed** fields. Lots of other fields could be stored which are never mentioned in the mapping
5. Please add an example to the docs where you pass multiple indices. So people can clearly see that the right form is `es_indexes = c("index1", "index2")`

mfrasco:
Here are my comments:
1. I agree that it catches a lot of extra meta information about the fields. Currently, I am hard-coding the keywords used to remove some of that information. This is not a preferable technique as those keywords might change with different ES versions. I'll try to figure out how to discriminate between real fields and meta information.
2. Sure, I was trying to match [this API](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-mapping.html). There is also an API to get the mapping for a particular field (so you don't need all of the information for the entire index or type). However, this API isn't useful if we drop the datatype column as you suggested.
3. All of the bad answers for datatype are associated with the extra meta information (store, analyzer, format), right? If I can get rid of those rows, we wouldn't have any bad answers. I think that the datatype is useful because it could tell you if a field was indexed as text for search or keyword for aggregation.
4. Sure, depending on what we do with 3, I'll rename the function and re-write the docs.
5. Of course.

jameslamb:
Fair point on number 3. Yeah, you're right that if you figure out the metadata part then the data types will make sense. I take it back...you should keep them.

Could we call the function like `get_fields` or `list_fields` or something in that direction? I think the term `mapping` will be unfamiliar to those who don't know ES super well.

Also...if/when you figure this out, please add your name to `DESCRIPTION` as an author 😃

mfrasco:
I was able to remove meta data by removing any rows that didn't end in "type". Also, I searched around in the elastic search docs and it seems like the format for the mapping API is stable. That is, the structure that I hard-coded isn't too brittle. Sub-fields of nested objects are always called properties and fields that have been indexed in multiple ways are always called fields.

r-pkg/tests/testthat/test-repo_characteristics.R
```diff
@@ -35,7 +35,7 @@ test_that('R CMD check should not return any unexpected errors, warnings, or not
     testthat::skip_on_cran()
     
     # Check the package
-    x <- devtools::check(pkg = '../../../uptasticsearch'
+    x <- devtools::check(pkg = '../../../r-pkg'
```
mfrasco: I think this is correct, right?

jameslamb: I'll pull and check. But yeah it looks right. I forgot to change this when we removed the intermediate `uptasticsearch` directory.

r-pkg/R/elasticsearch_eda_funs.R
```diff
@@ -114,3 +114,76 @@ get_counts <- function(field
...
+#' # get the mapping for all types in the ticket_sales and customers indexes
+#' mappingDT <- retrieve_mapping(es_host = "http://es.custdb.mycompany.com:9200"
+#'                               , es_indexes = c("ticket_sales", "indexes"))
+#' }
+get_fields <- function(es_host
+                       , es_indexes = '_all'
+) {
+    
+    # Input checking
+    url <- uptasticsearch:::.ValidateAndFormatHost(es_host)
```
jameslamb: You don't need to do this. `.ValidateAndFormatHost` from this package is guaranteed to be found first on the search path. Remove the `:::`

mfrasco: Ya. I did this so that I can run through the source code easily when testing. I just forgot to delete and didn't do R CMD check.

r-pkg/tests/testthat/test-repo_characteristics.R
mfrasco: Okay. I figured it out. If you run devtools::test(), it passes with this change. However, if you run devtools::check(), it does not because the working directory is different. We could set the working directory before this line.

jameslamb:
This is looking great! The more I see this code and play with it, the more convinced I am that this will be a cool and valuable feature to add.

A few more requests:

1. Can you change your regex to actually ensure that you are keeping only fields that _end_ in ".type"? So make it `'\\.type$'. When I tried the current state of this branch there were some metadata fields that were still getting caught. The dollar sign fixes that.
2. Because this function could be long-running (e.g. in the default case where you are hitting all indices), can you please add logging with `futile.logger` consistent with what we've done in `es_search`? 
3. In your `setnames` call, please specify an `old` and `new` argument. Calling that function without and `old` arg is equivalent to using numeric indices for column referencing, which makes me nervous.
4. I noticed in my testing that if you pass an alias to `es_indexes`, the resulting index in the data.table is the underlying ACTUAL index name (not the alias name). I'm trying to figure out how I feel about that. My gut says that it would be worth preserving the alias and returning that instead (seems more friendly for data scientists, who shouldn't have to know the details of aliasing). This would be easy to implement...you can get a table of aliases by hitting `hostname:port/_cat/_aliases`. @ngparas @austin3dickey what do you think?
5. Can you please use `data_type` instead of `datatype`? I know it's petty but it's more consistent with the style of the package (that everything external-facing is snake-case)
6. Add yourself to the `DESCRIPTION`!!

jameslamb:
Oh and re: your initial post, in case my last comment wasn't clear....this function ABSOLUTELY "has life". I want it. Please also add unit tests, mocking the requests with reads of some sample mappings stored in JSONs in `inst/testdata`

r-pkg/R/elasticsearch_eda_funs.R
```diff
@@ -114,3 +114,93 @@ get_counts <- function(field
...
+    
+    ########################## make the query ################################
+    futile.logger::flog.info(paste('Getting indexed fields for indexes:', indexes))
+    
+    result <- httr::GET(url = url)
+    httr::stop_for_status(result)
+    resultContent <- httr::content(result)
+    
+    ##################### return the flattened result #########################
+    return(.flatten_mapping(mapping = resultContent))
```
jameslamb: this is great. Love the new structure

mfrasco:
I addressed all of these concerns except for number four. The alias API doesn't return a json object. It returns a string. It looks like I can parse the string, but I wanted confirmation before I spent the time to do it. I also wanted to write a function to hit `hostname:port/_cat/indices&v`, but that has the same type of unstructured string as its output.

r-pkg/R/elasticsearch_eda_funs.R
mfrasco: Thanks. I had to do that in order to unit test. and then i realized that this was probably a better design anyway.

r-pkg/tests/testthat/test-elasticsearch_eda_funs.R
```diff
@@ -0,0 +1,63 @@
...
+                  expect_identical(mappingDT, expected)
+              }
+    )
+    
+    # works if multiple indexes are passed
+    test_that("get_fields should work if the mapping for multiple indexes are provided",
+              {
+                  test_json <- system.file("testdata", "two_index_mapping.json", package = "uptasticsearch")
+                  mapping <- jsonlite::fromJSON(txt = test_json)
+                  mappingDT <- uptasticsearch:::.flatten_mapping(mapping = mapping)
```
jameslamb: This is not actually testing the function. Can you please use `testthat::with_mock`? Mocking stuff:

```
`httr::GET` = function(...){return(TRUE)},
`httr::stop_for_status(...){return(invisible(NULL))},
`httr::content(...){return(paste0(readLines(test_json), collapse = ""))}`
```

r-pkg/R/elasticsearch_eda_funs.R
jameslamb: yeah I really like this. Very clean and readable

jameslamb:
For the alias thing...yeah I think you should try to parse that string. Sorry, I didn't realize the response would be so gross.

I got as far as this, but I'm sure there's a better way

```
x <- httr::GET(url = "http://hostname:port/_cat/aliases")
x <- gsub('- - -', '', x)
x <- strsplit(x, split = '\n')
x <- lapply(x, trimws)
x <- lapply(x, function(thing){gsub('\\s+', ' ', thing)})
```

r-pkg/R/elasticsearch_eda_funs.R
```diff
@@ -197,10 +218,47 @@ get_fields <- function(es_host
...
+    # construct the url to the alias endpoint
+    url <- paste0(es_host, '/_cat/aliases')
+    
+    # make the request
+    result <- httr::GET(url = url)
+    httr::stop_for_status(result)
+    resultContent <- httr::content(result)
+    
+    if (is.null(resultContent)) {
+        # there are no aliases in this Elasticsearch cluster
```
jameslamb: great catch, didn't even think about this

r-pkg/R/elasticsearch_eda_funs.R
```diff
@@ -163,14 +168,30 @@ get_fields <- function(es_host
     httr::stop_for_status(result)
     resultContent <- httr::content(result)
```
jameslamb: can you please explicitly state whether this is being parsed or you're just retrieving the text? i.e. make this `httr::content(result, as = 'parsed')`

r-pkg/R/elasticsearch_eda_funs.R
```diff
@@ -120,13 +120,17 @@ get_counts <- function(field
...
 #' @importFrom httr GET content stop_for_status
+#' @importFrom data.table := uniqueN
 #' @param es_host A string identifying an Elasticsearch host. This should be of
 #'                the form \code{[transfer_protocol][hostname]:[port]}. For example,
 #'                \code{'http://myindex.thing.com:9200'}.
 #' @param es_indexes A character vector that contains the names of indexes for
 #'                   which to get mappings. Default, is \code{'_all'}, which means
 #'                   get the mapping for all indexes. Names of indexes can be
 #'                   treated as regular expressions.
+#' @param use_alias A boolean flag that controls whether the true Elasticsearch
```
jameslamb: Kill off this flag. We don't need it. If people really want to get in the weeds and hit individual indexes instead of aliases, they can specify that in `es_indexes`. The way you've designed this function, it's already flexible to a mixing of aliases and actual index names.

I want to minimize the number of flags we have to maintain. and the associated forks in the code.

jameslamb: in other words....I'm saying just do the alias replacement by default. Fits in with our promise of doing things in a DS-friendly way that requires minimal knowledge of Elasticsearch

jameslamb:
Looks amazing, sir. Thanks for being patient with me. I'll test this tomorrow and let you know if I find anything.

Other than that, all I think this still needs is testing of `get_fields` with mocking. Let me know if you want me to give more details on `testthat::with_mock` or what I'm looking for.

Thanks for all your work!

mfrasco:
I'm a little confused as to what you want to do with `with_mock`. I could re-define get_fields to exclude the `httr` section. But I already tested the other components of that function. What did you have in mind?

jameslamb:
Sure, TL;DR right now you aren't testing the entire function. I'll give a more detail answer tonight

jameslamb:
image

jameslamb:
^ all of that code is currently uncovered by tests, because you're testing sub-components of the function and not the entire function. Ideally, in addition you would have tests of this form

```
test_that('get_fields works as expected', {
    testthat::with_mock(
        `httr::stop_for_status` = function(...){return(NULL)},
        `httr::GET` = function(...){return(NULL)}
        `httr::content` = function(...){return(readLines("sample_mapping.json"))},
        `uptasticsearch::.get_aliases` = function(...){someAliasDT},
        {
            outDT <- get_fields(es_host = "http://db.mycompany.com:9200"
                                , es_indexes = c("customers", "locations"))

            expect_true('data.table' %in% class(outDT))
            expect_true(nrow(outDT) > 0)
            expect_named(outDT, c('index', 'type', 'field', 'data_type'),
                         ignore.case = FALSE, ignore.order = TRUE)
        }
)})
```

jameslamb:
One other small thing...can you change `es_indexes` to `es_indices`? To be consistent with the language that Elasticsearch uses?

mfrasco:
I actually did think about this. It looks like ES 1.x used indexes and newer versions use indices. I think I chose indexes because I saw that it was used more outside of ES. But I agree that we should be consistent with ES.

r-pkg/R/elasticsearch_eda_funs.R
```diff
@@ -114,3 +114,145 @@ get_counts <- function(field
...
+#' @importFrom futile.logger flog.fatal flog.info
+#' @importFrom httr GET content stop_for_status
+#' @importFrom data.table := uniqueN
+#' @param es_host A string identifying an Elasticsearch host. This should be of
+#'                the form \code{[transfer_protocol][hostname]:[port]}. For example,
+#'                \code{'http://myindex.thing.com:9200'}.
+#' @param es_indices A character vector that contains the names of indices for
+#'                   which to get mappings. Default, is \code{'_all'}, which means
+#'                   get the mapping for all indices Names of indices can be
+#'                   treated as regular expressions.
```
austin3dickey: Could you please fix the punctuation in this param? (remove comma in 128, add period in 129)

r-pkg/R/elasticsearch_eda_funs.R
```diff
@@ -114,3 +114,145 @@ get_counts <- function(field
...
+    aliases <- stringr::str_split(aliasString, '\n')[[1]]
+    
+    # remove white space and only take the first two entries
+    aliases <- stringr::str_split(aliases, '\\s+')
+    aliases <- lapply(aliases, function(pair) pair[1:2])
+    
+    # create a data table from the resulting list
+    aliasDT <- data.table::as.data.table(matrix(unlist(aliases), byrow = TRUE, ncol = 2))
+    data.table::setnames(aliasDT, old = colnames(aliasDT), new = c('alias', 'index'))
+    return(aliasDT)
```
austin3dickey: I think this whole function can be replaced by
```
aliasDT <- data.table::data.table(read.table(text = alias_string, stringsAsFactors = FALSE))
return(aliasDT[, .(alias = V1, index = V2)])
```
It's twice as fast on my machine!

mfrasco:
@austin3dickey Thanks for the suggestion. It's much more straightforward. I checked that your method was robust enough to handle a case like this: https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-alias.html. This does add utils as a new dependency.

mfrasco:
Thanks Terry. I was meaning to do that.

terrytangyuan:
@mfrasco No problem. It's cool that Github now allows to edit the conflicts in the browser. Thank you for your continuous contributions!

mfrasco:
@ngparas Hey, do you have any questions about this PR?

PR #25, opened 2017-08-02
Fix bug when query return 0 hit

staftermath:
Fixed a bug in which `es_search` error out when query return 0 hit.
It now return `NULL`. A warning will be throw before returning `NULL`
Remade docs

### DIFF
--- a/r-pkg/DESCRIPTION
+++ b/r-pkg/DESCRIPTION
@@ -6,6 +6,7 @@ Authors@R: c(
     person("James", "Lamb", email = "james.lamb@uptake.com", role = c("aut", "cre")),
     person("Nick", "Paras", email = "nick.paras@uptake.com", role = c("aut")),
     person("Austin", "Dickey", email = "austin.dickey@uptake.com", role = c("aut")),
+    person("Weiwen", "Gu", email = "weiwen.gu@uptake.com", role = c("ctb")),
     person("Uptake Technologies Inc.", role = c("cph")))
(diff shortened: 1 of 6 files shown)


r-pkg/DESCRIPTION
```diff
@@ -6,6 +6,7 @@ Authors@R: c(
     person("James", "Lamb", email = "james.lamb@uptake.com", role = c("aut", "cre")),
     person("Nick", "Paras", email = "nick.paras@uptake.com", role = c("aut")),
     person("Austin", "Dickey", email = "austin.dickey@uptake.com", role = c("aut")),
+    person("Weiwen", "Gu", email = "weiwen.gu@uptake.com", role = c("ctb")),
```
jameslamb: 🎉

r-pkg/R/elasticsearch_parsers.R
```diff
@@ -804,6 +804,11 @@ es_search <- function(es_host
     
     # If we got everything possible, just return here
     hits_pulled <- length(firstResult[["hits"]][["hits"]])
+    
+    if (hits_pulled == 0) {
+      return(NULL)
```
jameslamb: I think we need to add a warning here. Something like "Your query was syntactically valid but matched 0 documents. Returning NULL."

Thoughts @austin3dickey ?

austin3dickey: Great idea.

PR #26, opened 2017-08-08
Fix dev install path in README and bump version to 0.0.2.999

jameslamb:


r-pkg/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: uptasticsearch
 Type: Package
 Title: Get Data Frame Representations of 'Elasticsearch' Results
-Version: 0.0.2
```
terrytangyuan: Actually 9999 is more commonly used!

jameslamb: ok can change it

PR #39, opened 2017-10-31
replaced httr:POST with httr::RETRY in elasticsearch_parses.R

ErhardMenker:


r-pkg/R/elasticsearch_parsers.R
```diff
@@ -1104,7 +1104,7 @@ es_search <- function(es_host
     }
     
     # Make request
-    result <- httr::POST(url = reqURL, body = query_body)
+    result <- httr::RETRY(url = reqURL, body = query_body)
```
jameslamb: Thank you for submitting a PR! However, this still needs some work:

1) See `?RETRY` for how that function works. They're not literally 1-to-1 swappable like this. You'll need to set some value for number of retries and provide an HTTP verb. You can test by trying to hit `"http://www.google.com"` with `httr::GET` and then seeing if you can get the same result with `httr::RETRY()`

2) Please add `RETRY` to the [roxygen docs](https://cran.r-project.org/web/packages/roxygen2/vignettes/roxygen2.html) for this function and regenerate them using `devtools::document`. You should see at least one `.Rd` file show up in the diff

3) This is not the only occurrence of `httr::POST` in the package. There are actually 4 of them. You can find them with `git grep` or with [find in files](https://support.rstudio.com/hc/en-us/articles/200710523-Navigating-Code) if you have RStudio.

ErhardMenker:
Hmmmmmmmmmmm...I need help on this.

1) What is a verb? The documentation isn't too helpful on this one and I'm
so unfamiliar about this that searching isn't helping too much. Is it just
a "GET?" Also is there a standard for the number of "retries?" I don't
understand what retrying does, if you can't access the web page isn't it
over?

2) I don't understand step 2.

jameslamb:
No problem, stick with it!

HTTP "verbs" help a web server to understand what you're trying to do when you send it a request. See here: https://robm.me.uk/web-development/2013/09/20/http-verbs.html

There are many reasons for a request to fail besides a URL not pointing to a valid resource. A web server could temporarily stop sending you data if you hit it too hard, your request could take too long to complete because of some transient disruption in your network...lots of thing could happen. Run `?httr::RETRY` in R and read all the documentation for that function. If you don't feel comfortable proposing a number of retries in this PR, just use 5.

`roxygen` is an R library for auto-generating LaTeX documentation using special commenting on R functions. See [my R class notes](https://jameslamb.github.io/teaching/mu_rprog/docs/code/Week5_Supplement.html) or Google around...there are tons of examples.

jameslamb:
We are going to decline this PR and open up the issue to other developers. Please come back and contribute in the future!

PR #40, opened 2017-11-13
Pkgdown hosting

jameslamb:
We previously kept the R version of `uptasticsearch` in an `r-pkg/` directory to separate it from an eventual Python version. This made sense for that use case, but going one level further in the directory structure meant we couldn't use turnkey GitHub pages hosting.

Since the Python version of the library does not seem to be imminent, in this PR I propose moving everything up a level so we can host our `pkgdown` docs on GitHub pages easily


r-pkg/.Rbuildignore
```diff
@@ -1,7 +0,0 @@
-^.*\.DS_Store$
-
-^docs$
-^_pkgdown\.yml$
-cran-comments.md
-tests/testthat/test-integration_tests.R
-inst/testdata/shakespeare_mapping.json
```
austin3dickey: Can you restore these lines in the new .Rbuildignore? I think they got lost

r-pkg/LICENSE
```diff
@@ -1,3 +0,0 @@
-YEAR: 2017
-COPYRIGHT HOLDER: Uptake Technologies Inc.
-ORGANIZATION: Uptake Technologies Inc.
```
austin3dickey: I'm a bit confused as to what's happening with the LICENSE. Looks like we're keeping around the BSD-3 clause LICENSE.MD in the top level and deleting this text file? But its language is still around in the pkgdown LICENSE.html? Perhaps we should consider just the BSD-3 clause instead of 
image

austin3dickey:
On your Github branch I don't see any links to a site or anything, is that expected? https://github.com/jameslamb/uptasticsearch/tree/pkgdown_hosting

r-pkg/.Rbuildignore
jameslamb: oh weird, I'm surprised that this got overwritten

I'm going to be unhappy if I learn that `pkgdown` overwrote this >.<

r-pkg/LICENSE
jameslamb: ugh what?? I didn't realize the two files were difference

yeah I messed this up. We probably need to keep both files

jameslamb:
@austin3dickey to your question...yeah, you wouldn't see the link on that branch

r-pkg/.Rbuildignore
jameslamb: ugh yeah I'm pretty sure `pkgdown` overwrote this file. Worst

r-pkg/LICENSE
jameslamb: We need to keep around both files. the language might have just made it into `pkgdown` because of the order I did things

austin3dickey:
I'm still a bit confused about the LICENSE; now we have a LICENSE and LICENSE.MD in the top level. The DESCRIPTION points to LICENSE while LICENSE.MD just seems to be the full BSD_3_clause template filled in.

[The CRAN instructions](https://cran.r-project.org/web/licenses/BSD_3_clause) seem unclear to me, but my interpretation is that we should remove LICENSE.MD.

austin3dickey:
(can't seem to reply to my previous comment)

Just got clarification from @jameslamb that we do need both LICENSE (for CRAN) and LICENSE.MD (for GitHub). That's the way to do it! 👍

.Rbuildignore
```diff
@@ -5,3 +5,6 @@
 cran-comments.md
 tests/testthat/test-integration_tests.R
 inst/testdata/shakespeare_mapping.json
+CONDUCT.md
+LICENSE.MD
+.travis.yml
```
austin3dickey: I'd suggest adding
- all the stuff in the .gitignore
- .gitignore itself
- ^ and $ and \ to match regex better

jameslamb: good call! will add them right now

jameslamb: added!

Issue #42, opened 2017-11-26
`unpack_nested_data` is much slower than `tidyr::unnest`


wdearden208:
From what I can see, `unpack_nest_data` is an order of magnitude slower than `tidyr::unnest`.

```
library(tidyverse)
library(microbenchmark)
library(uptasticsearch)
library(data.table)

n <- 1000

nested_df <-
  tibble(
    x = 1:n,
    y = rep(list(c("a", "b", "c")), n)
  )

nested_dt <- as.data.table(nested_df)

microbenchmark(
  unnest(nested_dt),
  unnest(nested_df),
  unpack_nested_data(nested_dt, col_to_unpack = "y"),
  times = 100
)
```

Since `unpack_nested_data` has the same functionality as `tidyr::unnest` but is limited to `data.table` and unnesting one column, it should be faster than `tidyr::unnest` in those cases.

Most of the computational cost is in the line `lapply(listDT, data.table::as.data.table)`. I wrote a version of the function which uses the basic idea of `tidyr::unnest` without the nonstandard evaulation and without the dplyr verbs (but still with purrr since that's already an imported package).

```
unpack_nested_data <- function(chomped_df, col_to_unpack)  {
  
    if (!("data.table" %in% class(chomped_df))) {
        msg <- "For unpack_nested_data, chomped_df must be a data.table"
        futile.logger::flog.fatal(msg)
        stop(msg)
    }
    if (".id" %in% names(chomped_df)) {
        msg <- "For unpack_nested_data, chomped_df cannot have a column named '.id'"
        futile.logger::flog.fatal(msg)
        stop(msg)
    }
    if (!("character" %in% class(col_to_unpack)) || length(col_to_unpack) != 
        1) {
        msg <- "For unpack_nested_data, col_to_unpack must be a character of length 1"
        futile.logger::flog.fatal(msg)
        stop(msg)
    }
    if (!(col_to_unpack %in% names(chomped_df))) {
        msg <- "For unpack_nested_data, col_to_unpack must be one of the column names"
        futile.logger::flog.fatal(msg)
        stop(msg)
    }

    outDT <- data.table::copy(chomped_df)
    listDT <- outDT[[col_to_unpack]]
    
    is_df <- purrr::map_lgl(listDT, is.data.frame)
    is_atomic <- purrr::map_lgl(listDT, purrr::is_atomic)
    if (all(is_df)) {
        newDT <- data.table::rbindlist(listDT)
[...]

Issue #29, opened 2017-08-28
Remove "old = names(DT)" from calls to data.table::setnames


austin3dickey:
A few times in our codebase, we have 

`data.table::setnames(someDT, old = names(someDT), new = new_names)`

This is actually _less_ safe than the following call with unnamed arguments:

`data.table::setnames(someDT, new_names)`

because the former will break when `someDT` has duplicate column names. It's not this package's job to break if there are duplicate names in a user's data.

kszela24:
Resolved with PR #46

PR #46, opened 2018-01-12
Remove "old = names(DT)" from calls to data.table::setnames

kszela24:
Found instances where data.table set names was called all of the names of the passed in DT.  Changed these instances to use the format of data.table::setnames(DT, newNames)


jameslamb:
Looks great! Thank you!

kszela24:
Reference #42

kszela24:
Reference #29

PR #48, opened 2018-01-12
Fix for data.table::setnames() potential bug

fparages:
Did not replace the ones in the shape of:

_data.table::setnames(outDT, names(outDT), paste0(aggNames, ".", names(outDT)))_

Because those would actually not break in the case of duplicated column names (old and new would have same length)


jameslamb:
Ah this was already addressed, sorry! Thank you for contributing though, come back soon!

PR #51, opened 2018-01-12
speed up unpack_nested_data #42

wdearden:
Reference #42


R/elasticsearch_parsers.R
```diff
@@ -402,46 +403,58 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+        msg <- "For unpack_nested_data, col_to_unpack must be all atomic vectors or all data frames"
+        futile.logger::flog.fatal(msg)
+        stop(msg)
+    }
     
-    # Rename unpacked column if it didn't get a name
+    # Create the unpacked data.table by replicating the originally unpacked
+    # columns by the number of rows in each entry in the original unpacked column
+    group_vars <- setdiff(names(chomped_df), c(names(newDT), col_to_unpack))
+    n <- pmax(purrr::map_int(listDT, NROW), 1)
```
jameslamb: TIL`pmax`

R/elasticsearch_parsers.R
```diff
@@ -391,7 +391,8 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
         futile.logger::flog.fatal(msg)
         stop(msg)
     }
-    if (!("character" %in% class(col_to_unpack)) || length(col_to_unpack) != 1) {
+    if (!("character" %in% class(col_to_unpack)) || length(col_to_unpack) != 
+        1) {
```
jameslamb: please put this `1) {` back on the line above, so the line ends with the opening `{`

R/elasticsearch_parsers.R
```diff
@@ -402,46 +403,58 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
-    oldIDs <- which(sapply(listDT, nrow) != 0)
-    listDT <- listDT[oldIDs]
-    
-    # Bind them together with an ID to match to the other data
-    newDT <- data.table::rbindlist(listDT, fill = TRUE, idcol = TRUE)
-    
-    # If we tried to unpack an empty column, fail
-    if (nrow(newDT) == 0) {
+    # Check for empty column
+    if (all(lengths(listDT) == 0)) {
```
jameslamb: 1) TIL `lengths()`
2) out of curiosity, what does this new version catch that `nrow(newDT)` didn't? It's not obvious to me

R/elasticsearch_parsers.R
```diff
@@ -402,46 +403,58 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
         msg <- "The column given to unpack_nested_data had no data in it."
         futile.logger::flog.fatal(msg)
         stop(msg)
     }
     
-    # Fix the ID because we may have removed some empty elements due to that bug
-    newDT[, .id := oldIDs[.id]]
+    listDT[lengths(listDT) == 0] <- NA
```
jameslamb: oh maybe I'm getting this. Will this catch the cases where you have a cell with value `list()`? I like this

R/elasticsearch_parsers.R
```diff
@@ -402,46 +403,58 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+                    x <- data.table::as.data.table(x)
+                }
+                x
+            }
+        )
+        newDT <- data.table::rbindlist(newDT, fill = TRUE)
+    } else if (all(is_atomic)) {
+        newDT <- data.table::as.data.table(unlist(listDT))
+    } else {
+        msg <- "For unpack_nested_data, col_to_unpack must be all atomic vectors or all data frames"
```
jameslamb: I'd like if we gave a more informative error message here. If I got this error as a user I wouldn't know what to do next.

Please do two things:
1) write a unit test using `expect_error()` that reaches this line
2) propose an error message format that might be able to give a user a bit more guidance on where to go next

R/elasticsearch_parsers.R
jameslamb: can you please use a more descriptive variable name than `n`?

R/elasticsearch_parsers.R
```diff
@@ -402,46 +403,58 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+        futile.logger::flog.fatal(msg)
+        stop(msg)
+    }
     
-    # Rename unpacked column if it didn't get a name
+    # Create the unpacked data.table by replicating the originally unpacked
+    # columns by the number of rows in each entry in the original unpacked column
+    group_vars <- setdiff(names(chomped_df), c(names(newDT), col_to_unpack))
+    n <- pmax(purrr::map_int(listDT, NROW), 1)
+    rest <- chomped_df[rep(1:nrow(chomped_df), n), ..group_vars, drop = FALSE]
```
jameslamb: it's not obvious to me, staring at this line, what it actually does. Can you please add a comment and choose a more informative name than `rest`?

jameslamb: what does `..groupvars`  do?

jameslamb: also what does `drop = FALSE` do?

R/elasticsearch_parsers.R
austin3dickey: This fails faster, I like it

jameslamb:
@wdearden208 I am going to approve this PR, looks great! Appreciate all the effort you put into this

However, I'm not going to merge. @austin3dickey you have the final 👍 / 👎 on this one

jameslamb:
@wdearden208 now that we merged that logging PR, you need to rebase this onto `master`

jameslamb:
I just updated this! It put me as a non-author committer, so you will still get credit for the contribution

R/elasticsearch_parsers.R
```diff
@@ -390,45 +390,64 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
     }
     
-    # Fix the ID because we may have removed some empty elements due to that bug
-    newDT[, .id := oldIDs[.id]]
+    listDT[lengths(listDT) == 0] <- NA
     
-    # Merge
-    outDT[, .id := .I]
-    outDT <- newDT[outDT, on = ".id"]
+    is_df <- purrr::map_lgl(listDT, is.data.frame)
```
austin3dickey: Could you please add `#' @importFrom purrr map_lgl` to the Roxygen of this function to be explicit?

R/elasticsearch_parsers.R
austin3dickey: Actually, I don't think this will fail if all the data.tables have 0 rows, just if they all have 0 columns. Was this intended?

R/elasticsearch_parsers.R
```diff
@@ -390,45 +390,64 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+            } 
+            x
+        }
+        newDT <- purrr::map(listDT, prep_row)
+        
+        newDT <- data.table::rbindlist(newDT, fill = TRUE)
+    } else {
+        msg <- paste0("Each row in column ", col_to_unpack, " must be a data frame or a vector.")
+        futile.logger::flog.fatal(msg)
+        stop(msg)
```
austin3dickey: You should use `log_fatal` here instead of these two lines

R/elasticsearch_parsers.R
```diff
@@ -390,45 +390,64 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+    # Bind packed column into one data.table
+    if (all(is_atomic)) {
+        newDT <- data.table::as.data.table(unlist(listDT))
+    } else if (all(is_df | is_atomic)) {
+        # If the packed column contains a mixture of data tables, we need to 
+        # to convert the atomic vectors to data.tables
+        
+        # Find column name to use for NA vectors
+        first_df <- min(which(is_df))
+        col_name <- names(listDT[[first_df]])[1]
```
austin3dickey: Hmm, this seems pretty arbitrary. I wonder if we should surface "which column should I put atomic data in, if it's a mixture of atomic and data.tables" to the user.

To that point, have you ever seen a payload that would result in `listDT` being a mixture of atomic and data.tables? I wonder if this whole section is too much complexity for a use case I've never seen.

R/elasticsearch_parsers.R
```diff
@@ -390,45 +390,64 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+        msg <- paste0("Each row in column ", col_to_unpack, " must be a data frame or a vector.")
+        futile.logger::flog.fatal(msg)
+        stop(msg)
+    }
     
-    # Remove the id column and the original column
-    outDT <- outDT[, !c(".id", col_to_unpack), with = FALSE]
+    # Create the unpacked data.table by replicating the originally unpacked
+    # columns by the number of rows in each entry in the original unpacked column
+    times_to_replicate <- pmax(purrr::map_int(listDT, NROW), 1)
```
austin3dickey: Why are you using pmax here? I thought `purrr::map_int()` is guaranteed to return one integer vector.

R/elasticsearch_parsers.R
```diff
@@ -390,45 +390,64 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+    # Replicate the rows of the data.table by entries of times_to_replicate but drop col_to_unpack
+    replicatedDT <- chomped_df[rep(1:nrow(chomped_df), times_to_replicate)]
+    replicatedDT[, col_to_unpack] <- NULL
```
austin3dickey: Makes sense, though perhaps you should test out my former strategy of using the built-in `idcol` argument of `rbindlist`. That's essentially replicating the logic `rep(1:nrow(chomped_df), times_to_replicate)` but I suspect it will be faster since it's data.table haha

R/elasticsearch_parsers.R
```diff
@@ -390,45 +390,64 @@ unpack_nested_data <- function(chomped_df, col_to_unpack) {
...
+        stop(msg)
+    }
     
-    # Remove the id column and the original column
-    outDT <- outDT[, !c(".id", col_to_unpack), with = FALSE]
+    # Create the unpacked data.table by replicating the originally unpacked
+    # columns by the number of rows in each entry in the original unpacked column
+    times_to_replicate <- pmax(purrr::map_int(listDT, NROW), 1)
+    # Replicate the rows of the data.table by entries of times_to_replicate but drop col_to_unpack
+    replicatedDT <- chomped_df[rep(1:nrow(chomped_df), times_to_replicate)]
```
austin3dickey: Makes sense, though perhaps you should test out my former strategy of using the built-in idcol argument of rbindlist. That's essentially replicating the logic rep(1:nrow(chomped_df), times_to_replicate) but I suspect it will be faster since it's data.table haha

austin3dickey: So I tried benchmarking a few versions of this function with a large `listDT` that was full of data.tables (no atomic):

1) the old function
2) the old function with `lapply(listDT, data.table::as.data.table)` commented out
3) this version

(3) was twice as fast as (1), but (2) was twice as fast as (3). I suspect the difference is all the copies you're making in this chunk.

wdearden: Yep, I think these changes lost the speedup. The main speedup was from dropping `data.table::as.data.table`.

R/elasticsearch_parsers.R
wdearden: Yes it will

R/elasticsearch_parsers.R
wdearden: Done

R/elasticsearch_parsers.R
wdearden: Done

R/elasticsearch_parsers.R
wdearden: Oh wait this is definitely just checking `nrow == 0` in a weird way. I'll fix it.

wdearden: Good point. I fixed it.

R/elasticsearch_parsers.R
wdearden: Ok I dropped the logic for handling mixed atomic/data.table columns but I still needed to grab the column name to give `NA` entries. I got an error if I didn't because in the "handle NAs" test it created a second `x` column with all `NA`.

R/elasticsearch_parsers.R
wdearden: I need to use `pmax` for the cases where the entry in `col_to_unpack` is empty, so the number of rows is 0. I need to `rep` it one time because I don't want to drop that row.

R/elasticsearch_parsers.R
wdearden: Try out this update's speed. For what it's worth I tried out this function where I used the same merge as the original function instead of `rep` and it ended up being about 20% slower. I think that makes sense because I'm still using a data.table `[`. So it shouldn't copy the data any more than a data.table merge would copy the data. It's basically going through the same algorithm as merging in the `data.table` with `.id` being the same as `times_to_replicate`. Furthermore, I think it makes sense this way since we don't have to deal with creating a `.id` column which could cause a problem if `chomped_df` has a `.id` column.

austin3dickey: Hey, thanks for your patience here. No idea why your build is failing. It may just be transient.

Just tested this update's speed on my test dataset again (which has 1000 data frames in the `col_to_unpack`, no atomic or lists or NAs). The numbers below are the median times each function took on my computer.
- Old function: 153 ms
- Old function with the `lapply(listDT, data.table::as.data.table)` commented out: 69 ms
- Your first version: 97 ms
- Your second version: 89 ms
- A hybrid: 17 ms

For the hybrid, I built off your second version. I changed the top to:
```
inDT <- data.table::copy(chomped_df)
listDT <- inDT[[col_to_unpack]]
inDT[, (col_to_unpack) := NULL]
inDT[, .id := .I]
```
then added the `idcol = TRUE` back to `rbindlist` and merged after `rbindlist` like so:
```
outDT <- inDT[newDT, on = ".id"]
outDT[, .id := NULL]
return(outDT)
```

So I still think there's value in researching `idcol` as an option. If you're worried about the `.id` column restriction, we could have that join column be a random UUID or something for safety. You can, for example, use `idcol = myUUIDvar` instead of `idcol = TRUE` to automatically name that ID column.

austin3dickey:
Whoops I commented on an outdated file.

austin3dickey:
For those keeping track at home, this final version outperforms the original function by 9x and `tidyr::unnest` by 2x!! (On my real-world test dataset)

MAJOR props to @wdearden . Thanks for the idea and execution.

Issue #53, opened 2018-03-15
Documentation for get_fields() mentions the wrong function


jameslamb:
https://uptakeopensource.github.io/uptasticsearch/reference/get_fields.html

These docs mention `retrieve_mapping` instead of the actual function name `get_fields()`. I believe we change the name mid-development but missed this in the docs

ankur-srivastava1:
I will be taking care of this.

ankur-srivastava1:
Just submitted a PR for this.

PR #56, opened 2018-04-18
Updated get_fields.html docs

Chronocook:
As per issue #53 
This is my second PR be nicer than the first James


jameslamb:
@Chronocook thank you! This PR is incomplete at the moment. Would you be willing to complete it?

You edited our `pkgdown` documentation, which gets auto-generated from the main docs. To complete this PR, you'd just have to make this same change here:

https://github.com/UptakeOpenSource/uptasticsearch/blob/master/R/elasticsearch_eda_funs.R#L130

and here:

https://github.com/UptakeOpenSource/uptasticsearch/blob/master/man/get_fields.Rd#L28

PR #62, opened 2018-05-10
Add assertions on input types

jameslamb:
This PR introduces `assertthat` as a dependency and adds a bunch of assertions on inputs. This will help in debugging and add extra protection against unintended queries being executed on users' ES clusters.


R/elasticsearch_parsers.R
```diff
@@ -628,6 +636,26 @@ es_search <- function(es_host
...
+        , assertthat::is.string(es_index)
+        , es_index != ""
+        , assertthat::is.string(query)
+        , query != ""
+        , assertthat::is.string(scroll)
+        , scroll != ""
+        , assertthat::is.number(max_hits)
+        , max_hits >= 0
+        , assertthat::is.number(n_cores)
+        , n_cores >= 1
```
austin3dickey: Could you change the n_cores assertions to `assertthat::is.count(n_cores)`? `is.count` checks to make sure it's a single positive integer.

Unrelatedly, I wish `is.count` had a flag or something to allow 0, because checking for a single nonnegative integer is pretty tricky right now.

R/elasticsearch_parsers.R
```diff
@@ -628,6 +636,26 @@ es_search <- function(es_host
...
+        , assertthat::is.flag(break_on_duplicates)
+        , assertthat::is.flag(ignore_scroll_restriction)
```
austin3dickey: May want to check that these args aren't NA. Unfortunately `assertthat::is.flag(NA)` is TRUE because NAs are logical by default.

R/elasticsearch_eda_funs.R
```diff
@@ -44,6 +45,24 @@ get_counts <- function(field
...
+        , assertthat::is.string(start_date)
+        , start_date != ""
+        , assertthat::is.string(end_date)
+        , end_date != ""
+        , assertthat::is.string(time_field)
+        , time_field != ""
+        , assertthat::is.string(use_na)
+        , use_na != ""
+        , assertthat::is.number(max_terms)
+        , max_terms > 0
```
austin3dickey: `is.count` would work here too.

R/elasticsearch_parsers.R
jameslamb: yeah good call. I actually don't mind that they force you to break it into two things. More explicit is better than controlling with flags!

R/elasticsearch_parsers.R
jameslamb: ah yep, good catch

jameslamb:
@austin3dickey I just made the changes you recommended. The build failing was just `R CMD CHECK` doing its thankless but important job! I had used `query` in a few places where the object is actually called `query_body`.

Should hopefully build 🤞

jameslamb:
@austin3dickey this is building now and I made the recommended changes, so gonna merge it up. This will look (in a Github email) like I "dismissed your review", but I actually did do all the things.

Kind of wish this had tasks like Bitbucket

jameslamb:
@austin3dickey ha joke's on me apparently I need one positive review no matter what to merge.

At your earliest convenience, no rush

R/elasticsearch_eda_funs.R
```diff
@@ -44,6 +45,24 @@ get_counts <- function(field
...
+        , assertthat::is.string(start_date)
+        , start_date != ""
+        , assertthat::is.string(end_date)
+        , end_date != ""
+        , assertthat::is.string(time_field)
+        , time_field != ""
+        , assertthat::is.string(use_na)
+        , use_na != ""
+        , assertthat::is.count(max_terms)
+        , max_terms > 0
```
austin3dickey: This is redundant - `is.count` ensures it's 1 or greater

R/elasticsearch_parsers.R
```diff
@@ -628,6 +636,28 @@ es_search <- function(es_host
...
+        assertthat::is.string(es_host)
+        , es_host != ""
+        , assertthat::is.string(es_index)
+        , es_index != ""
+        , assertthat::is.string(query_body)
+        , query_body != ""
+        , assertthat::is.string(scroll)
+        , scroll != ""
+        , assertthat::is.count(max_hits)
+        , max_hits >= 0
```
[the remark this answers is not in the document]
jameslamb: blegh oh right, now I understand. Ok I'm going to remove this one. `max_hits = 0` is a plausible value.

R/elasticsearch_eda_funs.R
jameslamb: whoops, yep will fix

R/elasticsearch_parsers.R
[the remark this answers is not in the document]
jameslamb: also I agree that it's unintuitive

jameslamb:
ugh I think I know what this is. Trying locally, I was seeing a bunch of errors about multiple types not being allowed. The shakespeare data has mixed types :/

image

jameslamb:
This is working now! Gonna ship it

Issue #36, opened 2017-09-11
Integration tests are very minimal right now


jameslamb:
`uptasticsearch` integration tests need a lot of love. We need to test all of the functions in the package on live data in Elasticsearch

jameslamb:
closing this now that #66 was merged

Issue #43, opened 2017-12-13
add ES 6 tests on Travis


jameslamb:
ES 6 GA is out in the world now: 

- https://www.elastic.co/blog/elasticsearch-6-0-0-released
- https://www.elastic.co/guide/en/elasticsearch/reference/6.0/index.html
- https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking-changes-6.0.html

I haven't done due diligence on ES 6 yet but would love if someone would take some time to get our tests running against it

Issue #33, opened 2017-09-07
Unit test coverage is not 100%


jameslamb:
Currently, unit tests cover only 59% of the lines of code in this project. Ideally, we would like every line in the project to be covered by tests.

jameslamb:
happy to report we have near-100% coverage now as of #66! Still leaving this open...we would love help covering the last few tough-to-hit cases

jameslamb:
coverage as of #72 

![image](https://user-images.githubusercontent.com/7608904/41582182-80f5ca6e-7366-11e8-90c7-9cc736671e8b.png)

^ note that those `get_fields` lines are actually covered...there are two scrolling strategies (one for new, one for legacy), and only one of the two gets covered in any single set of tests (on a single ES version). They're both covered because our CI stuff tests over many versions

jameslamb:
so our coverage is above 97%! Those last few lines would be real real hard to cover, but someone is welcome to take this up if they really want

jameslamb:
If you want to contribute on this, here's how you can check the test coverage:

1. Spin up Elasticsearch locally and seed it with data. This requires that you have `docker` running

```
./setup_local.sh 6.2
```

2. Install the package and run the R tests

```
make install_r
make coverage_r
```

You should see something like this in the terminal

![image](https://user-images.githubusercontent.com/7608904/67450665-c7cd1b80-f5e3-11e9-83da-3b74cc5f467a.png)

and a file `coverage.html` should open in your default web browser. That file shows detailed information on which files have untested code and which lines specifically are untested.

3. Add tests in `R-package/tests/testthat` and repeat step 2

4. When you're done, be sure to clean up your local environment

```
./cleanup_local.sh
```

PRs that add any test coverage are welcomed! You do not have to get to 100% in one shot.

PR #66, opened 2018-05-22
Add integration tests and make tests easier to develop

jameslamb:
This PR adds a bunch of integration tests and makes it easier to test locally.

* added creation of an empty index so we can test the "your query was valid but returned nothing" code
* wrapped all `return(NULL)` in `return(invisible(NULL))` to avoid phantom NULLs in logs
* changed strategy for scrolling requests. Scrolling queries are not working on ES6 right now. They changed ES6 to [this structure](https://www.elastic.co/guide/en/elasticsearch/reference/6.x/search-request-scroll.html).
* related to that...moved stuff needed for `scroll_url` down into `.keep_on_pullin` so that thing can figure out how to format scrolls. This is a cleaner place to handle that anyway
* added a bunch of integration tests. Including **an actual test on scrolling which we somehow never had before**
* added a `docker-compose.yml` and supporting gnarly bash scripts to make it easier to develop tests locally

Test coverage as of this PR:

image

I think it's important we get this PR reviewed and working before the next release. Addresses #33 , #36 , and to some extent #43


README.md
```diff
@@ -186,3 +186,23 @@ This is a fairly new project and, as the version number indicates, should be reg
 ### Auth Support <a name="authsupport"></a>
 
 `uptasticsearch` does not currently support queries with authentication. This will be added in future versions.
+
+### Running Tests Locally
+
+When developing on this package, you may want to run Elasticsearch locally to speed up the testing cycle. We've provided some gross bash scripts at the root of this repo to help!
```
austin3dickey: Nice

tests/testthat/test-integration.R
```diff
@@ -16,22 +16,181 @@ if (!identical(loggerOptions, list())){
 }
 futile.logger::flog.threshold(0)
 
-#--- 1. es_search
+#--- es_search
```
austin3dickey: Nice

tests/testthat/test-integration.R
```diff
@@ -16,22 +16,181 @@ if (!identical(loggerOptions, list())){
...
+        }, regexp = "When max_hits is not an exact multiple of size, it is possible to get a few more than max_hits results back")
+        expect_true(data.table::is.data.table(outDT))
+    })
+    
+    test_that("es_search works as expected for search requests that return nothing", {
+        testthat::skip_on_cran()
+        
+        # NOTE: Creating an intentionally empty index is the safest way to test
+        #       this functionality. Any other test would involve writing a query
+        #       and I want to avoid exposing our tests to changes in the query DSL
```
austin3dickey: Yeah I like this idea

austin3dickey:
Are you sure that scrolling is causing the problems? Diving into the failed builds, it looks like the errors could be unrelated to scrolling. Some erroring versions don't even use scrolling in the failed tests.

setup_es.sh
```diff
@@ -0,0 +1,41 @@
...
+         -X POST "http://${ES_HOST}/shakespeare/_bulk" \
+         -H 'Content-Type: application/json' \
+         --data-binary "@$filename"; 
+done
+
+# Add an intentionally empty index
+curl --silent \
+     -X PUT "http://${ES_HOST}/empty_index" \
+     -H 'Content-Type: application/json' \
+     -d @shakespeare_mapping.json
```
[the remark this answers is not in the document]
jameslamb: we talked but responding here for other viewers on the PR. This is empty in the sense that we never write any docs to it. Like every query you execute against it is going to return 0 docs.

This seemed like the safest way to get test coverage on the code that handles "you wrote a valid query but it happened to match nothing"

jameslamb:
@austin3dickey this is not ready for a second yet. Just want to let you know what I did tonight:

1. Just added a tiny bit of test data to the repo (instead of pulling all of that shakespeare data). ES6 doesn't allow multiple `_type`s for docs within one index. So I just intentionally saved a small sample of the data that's all a single type
2. The scrolling difference is definitely a thing...it just wasn't showing up in the logs because other changes I made to support ES 6.2 broke things (like types in indexes). So I implemented a switch that changes scroll request strategies.

Also I rebased to current master after merging #62 . Will keep working on this! I think the package will be in a more stable place after this PR finally gets working

jameslamb:
ok fixed SOME of the issues causing this to fail. Basically the bulk endpoint was rejecting my test data because of a weird ordering of my `curl` arguments or newlines in the data file "or something".

Now I need to fix the lingering test failures in legacy ES versions, which I _think_ are a combination of the cluster taking too slow to respond (gonna try adding a `_refresh` post) and slight changes in the aggs API.

updates to follow

jameslamb:
ok @austin3dickey Travis is working right now but I'm feeling confident that it's gonna build. I had to do a lot of crazy stuff today to get it working, but they're all good changes. Like stuff that was actually broken that this testing exposed. TL;DR **uptasticsearch as it stands on master is just barely working legacy versions and doesn't really work for ES6**

changes since you last reviewed:

- when you don't have any aliases set up for a cluster, in ES1.x `_cat/aliases` returns `"[]"`. In later versions, it returns a `NULL` or the empty string `""`. Changed our code in `get_fields()` to account for the `"[]"` case.
- Fixed a wild bug in `.major_version()`. a CRAZY error was causing `major_version()` to resolve to version information about the host machine. Like "x86-apple-darwin". That function takes an argument `version_string` but I was referencing `version` internally. `base::version` is available in R by default and gives information on the version of R you're running. FML. added some tests on `.major_version` and `.get_es_version` to catch this kind of stuff in the future
- Added a separate mapping file for ES5. In ES5, Elasticsearch changed the default behavior for how `fielddata` is handled. You can read more [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html). Also added that `fielddata` change to the mapping for ES6
- Added a `_refresh` call in test setup. Looking back, it doesn't seem like the absence of this was an issue, but it is still a thing we should do. It's defense against occasional random failing builds.
- Changed types in ES6 mapping from "keyword" to "text" so results would be the same as the default behavior of "string" type in legacy versions. Basically our aggs tests where failing because in legacy versions with "string" type, "king henry" would show up as the terms "king" and "henry" (separate entries) in a `terms` agg. With "keyword" type in ES6, they show up as a single thing.
- Fixed number-of-rows check for aggregation queries. I was checking for the wrong number of rows in the aggregation tests (should have been 4, not 12)
- upgraded all our ES testing versions to the last release on each major version (e.g. 5.0.2 instead of 5.0.0). No one should be ignoring bugfix releases.
- removed all the defaults on internal functions. Since we're passing values through from the higher functions to these, we should drop the defaults to ensure that we catch issues where parameters are not getting passed through. This actually caught something!!! We weren't passing `ignore_scroll_restriction` through to `.fetch_all`.
- updated `.Rbuildignore` to ignore all the new files that might be laying around as a result of local testing
- removed use of `docker-compose` and simplified the "how to test stuff locally" setup. Our tests all assume you're only running on version of ES, so `docker-compose` was overkill. See the updated README for the setup. I can confirm I used the new setup MANY times to test issues specific to one version of another and it was super easy to work with

I think that when we merge this PR, we'll be able to close #36 . There are some lines that are gonna be super tough to cover in tests, so we should leave #33 open for now. Our coverage after merging this will be 85%-ish, but remember that we aren't testing `get_counts` and that that function is going to be removed in future releases. So our coverage of core functionality is upwards of 95%

Let me know if you have other questions!

jameslamb:
HEY IT ACTUALLY BUILT

jameslamb:
Hey thanks @austin3dickey for looking it over!! Gonna merge it up

PR #69, opened 2018-05-29
Add deprecation warning on get_counts

jameslamb:
**Overview**

In this PR, I'd like to propose deprecating `get_counts`. This function is not really central to what `uptasticsearch` does and will require extra administrative burden to test. As a reminder, `get_counts` basically generates queries to check the counts of different values on a field within a period of time.

This is outside of our mission for a couple of reasons:

* assumes that you have time-series data
* writes query DSL inside the package
* only useful for a very specific type of EDA

IMHO we should remove this function from the package.

**Deprecation Strategy**

I propose accepting this PR so this warning goes out with the next release to CRAN (which we will attempt whenever #66 is building and gets merged).

After that release gets to CRAN, I'll submit a PR straight-up removing this function from `uptasticsearch`.

We will wait at least one month to do the next release, giving people time to uncover this deprecation warning.

IF any interest is shown in the function (via people creating `issues`), we can discuss keeping it.

### DIFF
--- a/R/elasticsearch_eda_funs.R
+++ b/R/elasticsearch_eda_funs.R
@@ -32,6 +32,7 @@
 #'                      , end_date = "now"
 #'                      , time_field = "dateTime")
 #' }
+#' @note \href{https://github.com/UptakeOpenSource/uptasticsearch/pull/69}{get_counts will be deprecated soon}
 get_counts <- function(field
                       , es_host
                       , es_index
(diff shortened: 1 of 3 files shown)


R/elasticsearch_eda_funs.R
```diff
@@ -32,6 +32,7 @@
 #'                      , end_date = "now"
 #'                      , time_field = "dateTime")
 #' }
+#' @references \href{https://github.com/UptakeOpenSource/uptasticsearch/pull/69}{get_counts will be deprecated soon}
```
austin3dickey: Perhaps change this to `@note`? That may make more sense than "References"

jameslamb: good call

jameslamb:
@austin3dickey should we deprecate `parse_date_time` too (with the same strategy)? That functionality is also not specific to Elasticsearch

austin3dickey:
I don't think so. That's a utility that parses data you get from Elasticsearch (timestamps of the form "2016-06-28T08:53:07Z") that's not replicable in base R (I don't think you can parse military time zone abbreviations) so it seems to me that it belongs

jataggart:
I like the deprecation strategy, but is there any way or need to make it more obvious? Maybe call out the deprecations in the README or NEWS documents...

jameslamb:
Got it building! All of the failed builds were from transient Travis issues. I just kept clicking rebuild until they resolved themselves

PR #71, opened 2018-06-10
release v0.3.0

jameslamb:
This PR adds a few more tweaks for us to push v0.3.0 to CRAN! See the diff on `NEWS.md` for the highlights.

I think we've added significant enough changes in the two months since v0.2.0 that a new release is warranted. Most notably support for Elasticsearch 6.x!

I would like #70 to make it onto this release if possible. Other than that, don't think anything is holding us up and we should be able to push to CRAN by the end of the week.


jameslamb:
cc @staftermath check out all the new goodness!

jameslamb:
oh weird this is failing with the same error that was reported in #58 

hmmmmmmmm

tests/testthat/test-integration.R
```diff
@@ -166,9 +166,26 @@ futile.logger::flog.threshold(0)
         expect_true(all(outDT[, doc_count > 0]))
     })
     
+    # We have tests on static empty results, but this test will catch
```
jameslamb: whoops looks like I included this in a rebase when I shouldn't have! This will just get fixed when we merge #70 and then I'll rebuild. Sorry for the confusion

jameslamb: like I accidentally committed the test but not the changes from #70 . I was just a fool in the way I did this...promise it will make sense when we merge that one

docs/docsearch.js
```diff
@@ -0,0 +1,85 @@
+$(function() {
```
austin3dickey: Whoa cool

jameslamb: ha yeah! this just came with `pkgdown::build_site` on the newest version

jameslamb:
I'd like #73 to make it out on this release as well, if it's not too controversial. Will update the NEWS.md and then try a CRAN submission once we merge that PR

jameslamb:
just submitted this for approval to CRAN! Will update the `cran-comments.md` and merge this if / when we get accepted.

jameslamb:
got accepted to CRAN. Merging this up!! Thanks @austin3dickey !

PR #72, opened 2018-06-11
remove get_counts

jameslamb:
DON'T MERGE YET.

This change will go out after we merge #71 and release v0.3.0 to CRAN. #69 added a deprecation warning on the `get_counts` function that will ship to CRAN with the next release. This PR, which will be the first on the way to 0.4.0 (_or maybe 1.0.0???_), removes `get_counts` from the package. Unfortunately this didn't allow us to clean up the dependencies at all but it will simplify the package and make the philosophy of `uptasticsearch` clearer.

There are no tests in the diff because we never had any, which is fun.


jameslamb:
#71 is merged and 0.3.0 is (out on CRAN!)[https://cran.r-project.org/web/packages/uptasticsearch/index.html]

gonna merge this up as soon as travis builds

jameslamb:
very very excited to merge this one and simplify the package. This is the first step on our path from 0.3.0 to 1.0.0!

PR #73, opened 2018-06-12
hardened handling of aliases

jameslamb:
Sorry @austin3dickey , one more that I want to go out on the v0.3.0 release.

We don't have any tests right now on how aliases are handled in `get_fields`. This PR addresses that.

One thing to call out is that having multiple aliases point to the same index is a totally valid thing, and our code previously couldn't handle that. This PR adds that support as well.

One other minor bugfix...`url` is a base function, so I changed our internal references to `es_url`. I noticed it because I was getting the dreaded `closure not subsettable` error while debugging.

### DIFF
--- a/R/get_fields.R
+++ b/R/get_fields.R
@@ -3,8 +3,9 @@
 #' @name get_fields
 #' @description For a given Elasticsearch index, return the mapping from field name
 #'              to data type for all indexed fields.
+#' @importFrom data.table := rbindlist uniqueN
 #' @importFrom httr add_headers content GET stop_for_status
-#' @importFrom data.table := uniqueN
+#' @importFrom purrr map2
 #' @param es_indices A character vector that contains the names of indices for
 #'                   which to get mappings. Default is \code{'_all'}, which means
 #'                   get the mapping for all indices. Names of indices can be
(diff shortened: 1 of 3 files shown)


jameslamb:
fun fact ES5.x onwards and legacy ES versions return alias results slightly differently. Added a fix to this PR. I'm glad we caught it before the release

R/get_fields.R
```diff
@@ -57,11 +58,36 @@ get_fields <- function(es_host
...
+                , .y = rawAliasDT[["alias"]]
+                , .f = function(idx_name, alias_name, mappingDT){
+                    tmpDT <- mappingDT[index == idx_name]
+                    tmpDT[, index := alias_name]
+                    return(tmpDT)
+                }
+                , mappingDT = mappingDT
+            )
+            , fill = TRUE
+        )
```
austin3dickey: I think there's a better way you could do this with a data.table merge but these DTs are small anyway so it probably doesn't matter

jameslamb: yeah they're pretty small and this makes it really explicit. This is an EDA function so super-super fast performance is less important than correctness, IMHO

ngparas:
lgtm

R/get_fields.R
```diff
@@ -130,15 +156,43 @@ get_fields <- function(es_host
         # there are no aliases in this Elasticsearch cluster
         return(invisible(NULL))
     } else {
-        return(.process_alias(alias_string = resultContent))
+        major_version <- .get_es_version(es_host)
+        process_alias <- switch(
+            major_version
```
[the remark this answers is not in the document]
jameslamb: fair point! I will throw that up in an issue

Issue #83, opened 2018-09-13
Python package


jameslamb:
After much discussion internally at Uptake, we have decided to add the Python version of `uptasticsearch` to this project!!!

Initial plan is to host the Python package here and provide documentation on how to install it from source. Whether or not we eventually push to PyPi has not been decided.

In anticipating of this, I've created `R` and `Python` labels for issues in this project.

This issue can be closed once `py-pkg` is created.

jameslamb:
this was closed as of #84

PR #84, opened 2018-09-21
Python Package Release

ngparas:
This PR includes the python package noted in #83 

It does have some integration testing that can be run locally, but that will likely require some changes to work with travis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ devtools::install_github("UptakeOpenSource/uptasticsearch")
 
 ### Python <a name="pythoninstallation"></a>
 
-To install the python package:
+This package is not currently available on PyPi. To build the development version from source, clone this repo, then :
 
 ```
 cd py-pkg
--- a/README.md
+++ b/README.md
@@ -39,7 +39,12 @@ devtools::install_github("UptakeOpenSource/uptasticsearch")
 
 ### Python <a name="pythoninstallation"></a>
 
-We plan to release a Python implementation of this functionality, but that is not available at this time. Check back often!
+To install the python package:
+
+```
+cd py-pkg
+pip install .
+```
 
 ## Usage Examples <a name="examples"></a>
 
--- a/py-pkg/Makefile
+++ b/py-pkg/Makefile
@@ -1,0 +1,20 @@
+
+install:
+	# Install uptasticsearch Python package
+	cd py-pkg && \
+	python setup.py install && \
+	cd ..
+	
+docs:
+	# Create sphinx rst files for every package and subpackage
+	cd py-pkg && \
+	sphinx-apidoc -f -e -o docs uptasticsearch && \
+	cd docs && \
+	make html
+
+integration-tests:
+	test_success=0 ;\
+	docker-compose build --no-cache ;\
+	docker-compose run uptasticsearch bash -c "sleep 15; cd dummy_data; sh load_dummy_data.sh; cd ..; pytest" || test_success=1 ;\
+	docker-compose down ;\
+	exit $$test_success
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -39,7 +39,12 @@ devtools::install_github("UptakeOpenSource/uptasticsearch")
 
 ### Python <a name="pythoninstallation"></a>
 
-We plan to release a Python implementation of this functionality, but that is not available at this time. Check back often!
+To install the python package:
```
jameslamb: Can you add language about PyPi to be explicit? Like

*This package is not currently available on PyPi. To build the development version of the package from source, clone the repo and then do the following:*

py-pkg/dummy_data/es_5_mapping.json
```diff
@@ -0,0 +1,26 @@
+{
```
jameslamb: totally cool to check all of these in now. Can you please create an issue (you can do it before we've even merged this, honestly) indicating that we need to unify the CI strategies for R and Python?

Here's an example (if you want to link it) of testing R and Python packages in different sub-builds within one travis build using their nifty `matrix` thing: https://github.com/dmlc/xgboost/blob/master/.travis.yml#L15

py-pkg/dummy_data/load_dummy_data.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/sh
```
jameslamb: in the issue about CI, can you explicitly call out all the files currently under `py-pkg` which are redundant with what's currently available for CI in this repo? Basically so whoever picks up the issue knows what stuff to go clean up.

py-pkg/tests/test_util.py
```diff
@@ -0,0 +1,97 @@
...
+        """raises ValueError w/o proper port
+        """
+        with pytest.raises(ValueError):
+            _format_es_url("es.cluster.com/")
+        with pytest.raises(ValueError):
+            _format_es_url("es.cluster.com")
+        assert isinstance(_format_es_url("http://es.com:9200"), str)
+        assert isinstance(_format_es_url("es.es.com:9200"), str)
+
+    def test_nonsense(self):
```
jameslamb: best tests

py-pkg/uptasticsearch/aggs_parsers.py
```diff
@@ -0,0 +1,171 @@
+
+import pandas as pd
+
+
+def _terms_agg_to_df(aggs_json):
```
jameslamb: can you please create another issue that talks about how the "supported aggs" table in the README needs to get updated to talk about Python?

py-pkg/Makefile
```diff
@@ -0,0 +1,20 @@
+
+install:
+	# Install uptasticsearch Python package
+	cd py-pkg && \
```
jameslamb: can you add one more issue around moving R into `r-pkg/` for consistency? That would include updating any documentation, Travis, etc. so definitely should be its own PR

py-pkg/dummy_data/es_5_mapping.json
ngparas: #85

py-pkg/dummy_data/load_dummy_data.sh
ngparas: updated #85

py-pkg/uptasticsearch/aggs_parsers.py
ngparas: #86

py-pkg/Makefile
ngparas: #87

jameslamb:
woo! Thanks @ngparas !

PR #91, opened 2018-09-30
Unify R and Python CI strategies (fixes #85)

### LINKED ISSUE #85: Unify CI for Python and R
The Python code now duplicates some of the integration testing and isn't set up with Travis. Per @jameslamb Here's an example of testing R and Python packages in different sub-builds within one travis build using their nifty matrix thing: https://github.com/dmlc/xgboost/blob/master/.travis.yml#L15

The redundant files are now under `py-pkg/dummy_data`, and the logic that needs to be moved to travis is in `py-pkg/Makefile` and `py-pkg/docker-compose.yml`

jameslamb:
When we merged the Python package as of #84 , we merged in all of its testing details as well. This PR looks to address the following problems:

* there is a great deal of redundancy between the R and Python testing stuff right now, esp. in how the Elasticsearch server is set up and seeded with data
* the python package is not currently tested on Travis

I expect this PR's initial build to fail as I figure out how to build R and Python packages together on Travis, but the end goal here is that both packages will be covered by CI from this point forward.


jameslamb:
Ok @austin3dickey @ngparas this is ready for review. I tried many configurations and couldn't get Travis to reuse an `env` matrix over a matrix of different-language builds, so I ended up manually specifying each build.

In total, this results in 28 builds (2 languages, 14 ES versions).  Currently we test on the head of each minor version. I have a preference for still doing this (since, for example, our package was working on 6.0.x but breaking on 6.2.x as of stuff I saw on #43 ). We are also only testing one version of R and Python.

I'm open to suggestions on the right fix of combination of language and ES versions, but think that's a concern for a follow-up PR.

The most important thing here is that now both packages have tests that don't care about version, and we test cross-version compatibility by running those same tests against many different ES versions.  This is different than the way @ngparas had originally set up the integration tests for Python.

.travis.yml
```diff
@@ -1,30 +1,289 @@
...
-  - ES_VERSION=5.0.2
-  - ES_VERSION=5.3.3
-  - ES_VERSION=5.4.3
-  - ES_VERSION=5.6.9
-  - ES_VERSION=6.0.1
-  - ES_VERSION=6.1.4
-  - ES_VERSION=6.2.4
+# Manually specifying each build configuration.
+# Would be better to figure out how to build a matrix with two
+# languages but share the ES_VERSION matrix across all builds.
```
austin3dickey: Wow this does suck. Oh well.

r-pkg/.Rbuildignore
```diff
@@ -54,3 +54,7 @@ vignettes/*\.pdf
 ^lib$
 ^sandbox$
 ^coverage.html$
+
+# Stuff
```
austin3dickey: Ha

.travis.yml
jameslamb: yeah idk what else to do. Once you need to do something as custom what we need, the limitations of Travis become clear.

PR #92, opened 2018-10-02
Handle NA return values from detectCores() (fixes #79)

### LINKED ISSUE #79: Weird error when Parallel::detectCores() returns NA
Error message for reference: `Error in if (!is.integerish(x)) return(FALSE) : missing value where TRUE/FALSE needed`.

drkarthi:
When detectCores() returns NA, the default value of ceiling(detectCores()/2) converts the NA into NA_real_. The previous way of checking using assertthat::is.count() cannot handle NA_real_.


drkarthi:
#79

r-pkg/R/es_search.R
```diff
@@ -100,6 +100,13 @@ es_search <- function(es_host
         )
         log_fatal(msg)
     }
+  
+    # assign 1 core by default, if the number of cores is NA
+    if (is.na(n_cores)){
```
jameslamb: could you make this a bit stronger? Now that I think about it, what if it returns `NULL` or some other crazy stuff?

Recommendation:

`if (!assertthat::is.count(n_cores))`

r-pkg/R/es_search.R
```diff
@@ -112,7 +119,6 @@ es_search <- function(es_host
         , assertthat::is.string(scroll)
         , scroll != ""
         , max_hits >= 0
-        , assertthat::is.count(n_cores)
```
jameslamb: this should actually still be here @drkarthi . It's a guarantee that _as of this point in the code_ `n_cores` is a count. Even though that `if` block above would make this always TRUE, I don't consider it dead or wasted code to have this assertion here.

Can you please reinstate it?

r-pkg/R/es_search.R
jameslamb: oh wow! Really good find!

Yes this might be a bug in `assertthat`. I do know that there have been a bunch of small bugfixes on that project that have not made it to Github yet. Let me see if I can reproduce from the dev version.

For now for this PR, I like your suggestion of the `||` thing

jameslamb: @drkarthi ok just checked, definitely a bug!!!

```
git clone https://github.com/hadley/assertthat.git
cd assertthat
R CMD INSTALL .
Rscript -e "assertthat::is.count(NA); assertthat::is.count(NA_real_)"
```

You should [create an issue](https://github.com/hadley/assertthat/issues) and then a PR to fix it! Believe the problem is in the implementation of `is.integerish()` (an internal function used by `assertthat`)

Again, great find!

drkarthi: Thank you, James!!! I've just created an issue in asserthat and made a commit with the changes for this PR!

drkarthi: https://github.com/hadley/assertthat/issues/54

jameslamb: wooo thank you!

jameslamb:
@drkarthi three of the builds on Travis failed for issues that look totally unrelated to your code. I just manually re-ran them.

Assuming they build successfully, I will approve and merge this.

Thank you for the help!

PR #94, opened 2018-10-11
added devtools to Travis stuff explicitly (fixes #93)

### LINKED ISSUE #93: Travis might be broken
image

This _could_ be related to the changes in #85 . I overrode some things that were previously being handled automagically by Travis. I wonder if that includes manually installing `devtools`? Still weird that #85 built though...

jameslamb:
In #85 , I overrode Jim Hester's defaults for building R packages on Travis (to make building R and Python in the same thing possible). That means I explicitly had to write out how to build, install, and test packages.

I make a few `devtools::` calls in there but we never explicitly install `devtools`. At least once (see the linked issue), this caused Travis to break.

I don't want to be at the whim of whatever configuration Travis boxes have. This PR adds `devtools` installation explicitly to avoid that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ matrix:
       cache: packages
       env: ES_VERSION=1.0.0
       install:
+        - Rscript -e "install.packages('devtools', repos = 'http://cran.rstudio.com')"
         - Rscript -e "devtools::install('r-pkg')"
       script:
         - R CMD build r-pkg/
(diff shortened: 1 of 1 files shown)


austin3dickey:
Looks like you need to fix the Python build

jameslamb:
Ha oh @austin3dickey I see what I did. I put R-specific stuff in something global. Fixing...

jameslamb:
Thanks @austin3dickey ! Yeah for sure. btw in the future you have full authority to smash merge

PR #134, opened 2018-12-01
[python] internalized some details

jameslamb:
I ran [doppel](https://github.com/jameslamb/doppel)  on `uptasticsearch` and it exposed some differences in the APIs that I want to address. I feel strongly that there's value in the Python and R APIs being identical.

In python there's no such thing as a truly "internal" object, but common convention is to use a leading `_`. In this PR, I propose using this form of "internalization" on many of the objects in the python version of `uptasticsearch`.

### DIFF
--- a/py-pkg/uptasticsearch/clients.py
+++ b/py-pkg/uptasticsearch/clients.py
@@ -6,10 +6,10 @@ from requests.adapters import HTTPAdapter
 import os
 
 from uptasticsearch.util import _format_es_url
-from uptasticsearch.util import convert_to_sec
+from uptasticsearch.util import _convert_to_sec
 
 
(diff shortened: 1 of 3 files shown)


ngparas:
I'm not sure I agree that interfaces need to be identical. I'm all for the main things being equivalent, but I think there should probably be some room to adapt to the conventions of the language. The client objects feel natural in python code. Would you be open to having the things imported from the top level match, and then have the factory still available in `uptasticsearch.clients` rather than "private"?

jameslamb:
@ngparas yep makes sense to me

jameslamb:
this was well-intentioned but I think I'm going to close it for now. To be revisited in the future, maybe

PR #136, opened 2019-01-28
Removed closeAllConnections from tests (fixes #135)

jameslamb:
This should get CRAN off our backs. If you approve I'll cut a release and push it, but let's keep the PR open until we get accepted.

PR #138, opened 2019-01-30
changed strategy for dropping duplicates (fixes #137)

### LINKED ISSUE #137: unique()-ing records breaks es_search() as of data.table 1.12.0
Earlier today @fparages noted that `uptasticsearch::es_search()` was throwing an error like this:

> Error in forderv(x, by = by, sort = FALSE, retGrp = TRUE) :
> Column 2 of by= (2) is type 'list', not yet supported

We tracked this down to only occurring when the result had `hits` where at least one field was an array. In Felipe's case, this was occurring when he use sorting and every hit then came back with this in it:

```
"sort": [
    "all",
    "some_id"
]
```

After some investigation tonight, I discovered Rdatatable/data.table#3332. Basically running `unique(someDT)` when `someDT` contains any columns that are lists will throw an error as of `data.table` 1.12.0 (released two weeks ago).

jameslamb:
This addresses #137 . To be honest, we should have been unique-ing by `_id` anyway.

I feel ok about this change because you can't actually avoid `_id` being a part of each hit (AFAIK), even if you do the `"source": {"include"` trick.

### DIFF
--- a/.ci/before_install.sh
+++ b/.ci/before_install.sh
@@ -4,7 +4,7 @@ set -e
 
 # If language: r,
 # install these testing packages we need
-if [ -z "$TRAVIS_R_VERSION" ];
+if [[ "$TASK" == "rpkg" ]];
 then
   Rscript -e "install.packages(c('devtools', 'knitr', 'testthat', 'rmarkdown'), repos = 'http://cran.rstudio.com')"
 fi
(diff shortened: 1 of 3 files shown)


jameslamb:
don't merge this yet. If CRAN rejects us again on #136 , let's just get this into the next submission.

If not, then I'll bump the version on here to 0.3.2

PR #151, opened 2019-06-04
fixed Travis setup (fixes #149)

### LINKED ISSUE #149: Python builds in CI are broken
Got a notification recently from Travis (thanks cron feature!!) telling me that our builds are broken.

The error looks to be a change in the environment:

```
Package oracle-java8-set-default is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  oracle-java9-set-default
E: Package 'oracle-java8-set-default' has no installation candidate
apt-get.diagnostics
apt-get install failed
```

Fixing this might be as simple as upgrading to java9 (https://github.com/UptakeOpenSource/uptasticsearch/blob/master/.travis.yml#L6) but I don't know if that will make it harder for us to run legacy versions of ES.

PRs welcomed!

jameslamb:
Trying to fix Travis since we were (fairly!) called out on [this blog](https://blog.panoply.io/the-13-best-etls-for-r-programming) for letting it slip.


jameslamb:
woo it works! @ngparas could you take a look?

PR #161, opened 2019-08-22
support for ES 7 (fixes #155)

jameslamb:
This pull requests aims to add support in `uptasticsearch` for Elasticsearch 7.x. See the linked issue and changes to NEWS.md for details on what has changed in ES7.x.

This PR's scope is limited to "get `uptasticsearch` code working with Es7.x". It does not include taking advantage of any ES7-specific _features_ like new types of aggregations.

Opening this as a draft PR as it currently only addresses the R side. The Python side needs to be updated n this PR as well.

PR #163, opened 2019-08-30
simplified CI setup

jameslamb:
Looking at this project for the first time in a bit while working on #161 , I found that there are some places where can reduce duplication. In this PR, I want to propose some reorganization that will removed opportunities for error and make the exact amount of complexity (e.g. all the ES7-specific exceptions in test setup) more obvious.

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -12,5 +12,5 @@ fi
 
 if [[ "$TASK" == "pypkg" ]]; then
   PY_PACKAGE_DIR=$(pwd)/py-pkg
-  pip install ${R_PACKAGE_DIR}
+  pip install ${PY_PACKAGE_DIR}
 fi
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# failure is a natural part of life
+set -e
+
+if [[ "$TASK" == "rpkg" ]]; then
+  R_PACKAGE_DIR=$(pwd)/r-pkg
+  R CMD INSTALL \
+    --clean \
+    ${R_PACKAGE_DIR}
+fi
+
+if [[ "$TASK" == "pypkg" ]]; then
+  PY_PACKAGE_DIR=$(pwd)/py-pkg
+  pip install ${R_PACKAGE_DIR}
+fi
--- a/.ci/seed_es_on_travis.sh
+++ b/.ci/seed_es_on_travis.sh
@@ -96,13 +96,13 @@ case "$ES_VERSION" in
     
     "6.1.4")
       export ES_VERSION=6.1.4;
-      export MAPPING_FILE=${ES5_MAPPING_FILE};
+      export MAPPING_FILE=${ES6_MAPPING_FILE};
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -32,408 +41,156 @@ matrix:
...
         - ES_VERSION=6.2.4
         - TASK=rpkg
-      install:
-        - Rscript -e "devtools::install('r-pkg')"
-      script:
-        - R CMD build r-pkg/
-        - R CMD check *.tar.gz
       after_success:
-        - Rscript -e 'install.packages("covr"); Sys.setenv(NOT_CRAN = "true"); covr::codecov("r-pkg/")'
+        - .ci/report_to_covr.sh
```
austin3dickey: Maybe I'm misunderstanding the diff, but is there a reason you're only running the covr report on 6.2.4? Is it because you only want to run it once?

.ci/install.sh
```diff
@@ -0,0 +1,16 @@
...
+if [[ "$TASK" == "rpkg" ]]; then
+  R_PACKAGE_DIR=$(pwd)/r-pkg
+  R CMD INSTALL \
+    --clean \
+    ${R_PACKAGE_DIR}
+fi
+
+if [[ "$TASK" == "pypkg" ]]; then
+  PY_PACKAGE_DIR=$(pwd)/py-pkg
+  pip install ${R_PACKAGE_DIR}
```
austin3dickey: This should be `PY_PACKAGE_DIR`

.travis.yml
jameslamb: Yeah it only needs to run once. So right now we're using `testthat` to run the tests, and then just reporting to `codecov` (an external service) one time to power the CI checks around coverage and the badge in the README.

Eventually I'd like to just use `covr` to run tests (like you and I have done in other projects) but didn't want to pull that into this PR.

.ci/install.sh
jameslamb: oh good catch

jameslamb:
> Looks great! I saw that 2 R builds failed but I suspect it was transient?

I'm not sure if it's transient. Lemme try pushing the other changes I have an we'll see. If 6.1.x and 6.2.x fail for BOTH R and Python, we'll know I got the logic for pulling and starting up ES wrong.

.ci/seed_es_on_travis.sh
```diff
@@ -0,0 +1,152 @@
...
+    
+    "6.1.4")
+      export ES_VERSION=6.1.4;
+      export MAPPING_FILE=${ES5_MAPPING_FILE};
+      export ES_BINARY_URL="${ES5PLUS_ARCHIVE}/elasticsearch-$ES_VERSION.deb"
+      ;;
+      
+    "6.2.4")
+      export ES_VERSION=6.2.4;
+      export MAPPING_FILE=${ES5_MAPPING_FILE};
```
jameslamb: ah here's the error (probably?), I used the wrong mapping file.

Python tests didn't fail because there isn't a `get_fields()` in the Python package yet. Created #164 to document it.

jameslamb:
Ok this is passing! @austin3dickey whenever you have time, could you take one more look?

PR #174, opened 2019-09-25
added better type checks

jameslamb:
In this PR, I improved some of our assertions in the R package.

So e.g. `data.table::is.data.table(thingDT)` is preferred to `"data.table" %in% class(thingDT)`.


r-pkg/R/es_search.R
```diff
@@ -568,15 +568,15 @@ es_search <- function(es_host
...
+    if (! is.character(es_host)){
+        msg <- paste0("es_host should be a string. You gave an object of type"
                       , paste0(class(es_host), collapse = '/'))
         log_fatal(msg)
     }
 
     # [2] es_host is length 1
     if (! length(es_host) == 1){
-        msg <- paste0("es_host should be length 1!"
+        msg <- paste0("es_host should be length 1."
```
austin3dickey: haha

jameslamb: so important

jameslamb: it felt like we were being too aggressive

jameslamb:
> Looks good! All the R tests are failing though. Travis doesn't print the tracebacks so it's hard to tell why.

I'm sure I messed up something silly. Will take a look

jameslamb:
> > Looks good! All the R tests are failing though. Travis doesn't print the tracebacks so it's hard to tell why.
> 
> I'm sure I messed up something silly. Will take a look

1. I hate the way that commenting works on GitHub, where sometimes you can thread and sometimes you have to quote-reply
2. When I ran the tests locally, I was able to get the same tests to fail by not copying the test data into `r-pkg/inst/testdata`. I'm going to kick off a build where I directly run the tests (instead of `R CMD CHECK`) so maybe we can get more logs

r-pkg/R/es_search.R
```diff
@@ -85,7 +85,7 @@ es_search <- function(es_host
 ){
 
     # Check if this is an aggs or straight-up search query
-    if (length(query_body) > 1 || ! "character" %in% class(query_body)){
+    if (assertthat::is.string(query_body)){
```
jameslamb: this assertion was the problem with Travis! I should have made it `!is.string()` and I made it `is.string()`. Good for us for having good tests!

jameslamb: yeah this is working now 😄

PR #181, opened 2019-10-24
fix: Add missing '$' in substitution to get the major version

luisgdelahozl:


jameslamb:
sorry @luisgdelahozl  I saw #182 first and merged it, so this PR isn't necessary any mroe. Feel free to pick up another issue!

Also...I strongly recommend you get in the habit of adding `(fixes #the-issue-number)` to the end of PR titles. That way, when you PR is merged the issue it addresses will automatically be closed. It also builds a link between the PR and the issue so reviewers have context for why you are proposing a change.

luisgdelahozl:
It is ok. Thanks for the feedback.

PR #184, opened 2019-10-24
Fixes #178 - multi-line blocks in roxygen documentation should be ali…

### LINKED ISSUE #178: multi-line blocks in roxygen documentation should be aligned to the first non-tag word
currently, the R package has many `roxygen2` documentation blocks like this:

image

It would be better to standardize on aligning all lines to the first non-tag character, like this

image

To complete this issue, go through all of the files in `R-package/R/` and look for places where we have multi-line roxygen blocks. When changing the text, keep lines to under 100 characters in length.

chrsblck:
…gned to the first non-tag word

### DIFF
--- a/r-pkg/R/unpack_nested_data.R
+++ b/r-pkg/R/unpack_nested_data.R
@@ -1,19 +1,18 @@
 #' @title Unpack a nested data.table
 #' @name unpack_nested_data
 #' @description After calling a \code{chomp_*} function or \code{es_search}, if
-#'   you had a nested array in the JSON, its corresponding column in the
-#'   resulting data.table is a data.frame itself (or a list of vectors). This
-#'   function expands that nested column out, adding its data to the original
-#'   data.table, and duplicating metadata down the rows as necessary.
+#'              you had a nested array in the JSON, its corresponding column in the
+#'              resulting data.table is a data.frame itself (or a list of vectors). This
+#'              function expands that nested column out, adding its data to the original
+#'              data.table, and duplicating metadata down the rows as necessary.
 #'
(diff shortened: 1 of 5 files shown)


chrsblck:
@jameslamb Should `\dontrun{` be moved to the line below?

https://github.com/uptake/uptasticsearch/blob/a3a50fc1578f46273828f5a78a8e9bfa106d42a2/r-pkg/R/get_fields.R#L17

r-pkg/R/unpack_nested_data.R
```diff
@@ -1,19 +1,19 @@
...
-#'   input data.table is unmodified.
+#'              This is a side-effect-free function: it returns a new data.table and the
+#'              input data.table is unmodified.
 #' @importFrom data.table as.data.table copy is.data.table rbindlist setnames
 #' @importFrom purrr map_if map_lgl map_int
 #' @export
 #' @param chomped_df a data.table
 #' @param col_to_unpack a character vector of length one: the column name to
-#'   unpack
+#'                      unpack
```
jameslamb: for this one, could you move "unpack" up to the end of the preceding line? I think it was too aggressive for us to put it on its own line

jameslamb:
> @jameslamb Should `\dontrun{` be moved to the line below?
> 
> https://github.com/uptake/uptasticsearch/blob/a3a50fc1578f46273828f5a78a8e9bfa106d42a2/r-pkg/R/get_fields.R#L17

yes please, good catch

chrsblck:
@jameslamb Addressed both comments. I accidentally squashed the update too soon (bad habit). Sorry!

PR #193, opened 2019-10-24
Add lifecycle badges to README (fixes #188)

### LINKED ISSUE #188: Add lifecycle badges to README
I think this project's README should include lifecycle badges.

The Tidyverse folks have introduced [these lifecycle stages](https://www.tidyverse.org/lifecycle/) and associated badges.

I think the `uptasticsearch` R package is probably in the probably in the Maturing phase:

> In a stable package, we are largely happy with the API, and major changes are unlikely. This means that the API will generally evolve by adding new functions and new arguments; we will avoid removing arguments or changing the meaning of existing arguments.

> If API breaking change are needed, they will occur gradually. To begin with, the function or argument will be deprecated; it will continue to work but will emit an message informing you of the change. Next, typically after at least a year, the message will be transformed to an error

And the Python package is probably in the Dormant phase:

> A dormant package is not completed, but is not currently under active development. We plan to return to it in the future.

To complete this issue, add lifecycle badges for the Python and R packages to README.md. Use your best judgment for where to place those badges and how to make it clear which applies to Python vs. R...we can discuss that design choice on the PR.

See the [pkgnet README](https://github.com/uptake/pkgnet/blob/master/README.md) for an example of how to use these badges.

varadpoddar:


varadpoddar:
> Thanks for contributing! I have two requested changes:
> 
> 1. could you change the color for `dormant` to orange? We should use color to visually distinguish them
> 2. can you please add an explicit reference in the PR title? i.e. `Add lifecycle badges to README (fixes #188)`. That "fixes" is important to get GitHub to automatically close the issue when this PR gets merged.

1. Done
2. Done

PR #199, opened 2019-10-25
More syntax highlighting (fixes #196)

### LINKED ISSUE #196: explicitly add language tags to all markdown code blocks
Currently, code blocks in `README.md` use generic markdown (\`\`\`), which means we're missing out on some syntax highlighting!

To close this issue, edit `README.md` so that each code block indicates the language (e.g. \`\`\`r). See [here](https://raw.githubusercontent.com/willdebras/ctar/master/README.md) as an example.

you can use:

* python: \`\`\`python
* R: \`\`\`r
* shell scripts: \`\`\`shell

irfansener:


README.md
```diff
@@ -53,7 +53,7 @@ devtools::install_github(
 
 This package is not currently available on PyPi. To build the development version from source, clone this repo, then :
 
-```
+``` python
```
hpedrorodrigues: I think there should be `shell` or maybe `bash`.

irfansener: Yes you are right. I am changed it

jameslamb: Thanks for the comment @hpedrorodrigues! Yes, it should be `shell`. GitHub's markdown renderer, like many others, uses `highlight.js`. Here are all the supported languages, if you're curious: https://highlightjs.org/static/demo/

PR #205, opened 2020-03-09
FIX: replaces all `ES` with `Elasticsearch` (fixes #202)

### LINKED ISSUE #202: Replace use of 'ES' with 'Elasticsearch' everywhere in the repo
Everywhere other than in actual variable names in the code, I think we should standardize on `Elasticsearch` instead of `ES`. This will make documentation in the project a bit clearer, and might increase the relevance of this project to searches (on GitHub and through web search engines) for things like "Elasticsearch R client".

bernardbeckerman:
Since the comment syntax is very consistent, it was straightforward to do a sed replace with

```for file in `grep -Rl "\WES\W" *`; do sed -i "s@\([^a-zA-Z0-9]\)ES\([^a-zA-Z0-9]\)@\1Elasticsearch\2@g" $file; done```

Note that the above does not replace instances of the form `"ES6.x"` for example. Would be happy to add that on to this PR if desired.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,7 +109,7 @@ You may notice that this strategy means that `uptasticsearch` is tested for back
 
 When developing on this package, you may want to run Elasticsearch locally to speed up the testing cycle. We've provided some gross bash scripts at the root of this repo to help!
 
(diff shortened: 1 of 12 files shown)


bernardbeckerman:
@jameslamb updated per your request then squashed

PR #207, opened 2020-04-19
added linting on R code (fixes #179)

### LINKED ISSUE #179: Add linting on R code
We should use a linter to somewhat-standardize the style of the R code in this project. For R code, I'm fond of the `covr` package.

To close this issue, introduce linting in the tests run on Travis. This issue will involve reading our `.travis.yml` and the scripts in `.ci/` to get an understanding of how `uptasticsearch`'s Continuous Integration (CI) setup.

I recommend taking an approach similar to the one in microsoft/LightGBM#2437, but it's up to you (the contributor) to propose how you want to handle it.

This is a medium-difficulty issue, but one you can do without understanding anything about Elasticsearch or `uptasticsearch`'s R code.

jameslamb:
This PR introduces a linter for the R package, using  `{lintr}`.

### DIFF
--- a/.ci/lint_r_code.R
+++ b/.ci/lint_r_code.R
@@ -1,0 +1,109 @@
+
+library(lintr)
+
+args <- commandArgs(
+    trailingOnly = TRUE
+)
+SOURCE_DIR <- args[[1L]]
+
+FILES_TO_LINT <- list.files(
+    path = SOURCE_DIR
+    , pattern = "\\.r$"
+    , all.files = TRUE
+    , ignore.case = TRUE
+    , full.names = TRUE
+    , recursive = TRUE
+    , include.dirs = FALSE
+)
+
+# text to use for pipe operators from packages like 'magrittr'
+pipe_text <- paste0(
+    "For consistency and the sake of being explicit, this project's code "
+    , "does not use the pipe operator."
+)
+
+# text to use for functions that should only be called interactively
+interactive_text <- paste0(
+    "Functions like '?', 'help', and 'install.packages()' should only be used "
+    , "interactively, not in package code."
+)
+
+LINTERS_TO_USE <- list(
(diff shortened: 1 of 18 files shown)


.ci/lint_r_code.R
```diff
@@ -0,0 +1,109 @@
+
+library(lintr)
+
+args <- commandArgs(
```
austin3dickey, line 4: TIL `commandArgs`. Always better to use base R and reduce dependencies if it's easy. :)

r-pkg/R/es_search.R
```diff
@@ -737,8 +737,8 @@ es_search <- function(es_host
 # [examples]
 # \dontrun{
 # #--- Example 1: Basic Usage ---#
-# .ConvertToSec('1h') # returns 60*60 = 3600
-# .ConvertToSec('15m') # returns 60*15 = 900
+# .ConvertToSec('1h') # returns 60 * 60 = 3600
+# .ConvertToSec('15m') # returns 60 * 15 = 900
```
austin3dickey, line 741: interesting, it lints comments too!

.ci/test.sh
```diff
@@ -4,6 +4,7 @@
 set -e
 
 if [[ "$TASK" == "rpkg" ]]; then
+  Rcript .ci/lint_r_code.R $(pwd)
```
austin3dickey, line 7: I think you meant "Rcrypt" ☠️ 
```suggestion
  Rscript .ci/lint_r_code.R $(pwd)
```

r-pkg/R/es_search.R
jameslamb: ha it doesn't, but when I did find-in-all-files for `*6` these  showed up and I thought "eh might as well"

.ci/test.sh
jameslamb: ha yup that would do it

PR #211, opened 2020-05-01
[ci] remove devtools and fix small docs stuff

jameslamb:
I'd like to propose removing `devtools` from this project. We're installing it in CI (which is expensive because it has a lot of dependencies) but I don't think we're actually using it!

We have it in our docs for `install_github()`, but that has since [moved to `remotes`](https://cran.r-project.org/web/packages/remotes/readme/README.html).

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -11,7 +11,7 @@ export JAVA_APT_PKG="oracle-java8-set-default"
 # install these testing packages we need
 if [[ "$TASK" == "rpkg" ]];
 then
-  Rscript -e "install.packages(c('data.table', 'futile.logger', 'knitr', 'lintr', 'testthat', 'rmarkdown', 'uuid'), repos = 'http://cran.rstudio.com')"
+  Rscript -e "install.packages(c('assertthat', 'covr', 'data.table', 'futile.logger', 'httr', 'jsonlite', 'knitr', 'lintr', 'purrr', 'rmarkdown', 'stringr', 'testthat', 'uuid'), repos = 'http://cran.rstudio.com')"
   cp test-data/* r-pkg/inst/testdata/
 fi
 
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -11,7 +11,7 @@ export JAVA_APT_PKG="oracle-java8-set-default"
 # install these testing packages we need
 if [[ "$TASK" == "rpkg" ]];
 then
-  Rscript -e "install.packages(c('data.table', 'devtools', 'futile.logger', 'knitr', 'lintr', 'testthat', 'rmarkdown', 'uuid'), repos = 'http://cran.rstudio.com')"
+  Rscript -e "install.packages(c('data.table', 'futile.logger', 'knitr', 'lintr', 'testthat', 'rmarkdown', 'uuid'), repos = 'http://cran.rstudio.com')"
   cp test-data/* r-pkg/inst/testdata/
 fi
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -129,7 +129,7 @@ make coverage_r
 
 ### Checking code style
 
-The R package's code style is tested with `{lintr}`. To check the code locally, r un the following
+The R package's code style is tested with `{lintr}`. To check the code locally, run the following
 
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -147,13 +147,13 @@ This is a manual process, with the following steps.
...
 
-Add a section for this release to `NEWS.md`.  This file details the new features, changes, and bug fixes that occured since the last version.  
+Add a section for this release to `NEWS.md`.  This file details the new features, changes, and bug fixes that occurred since the last version.  
 
 Add a section for this release to `cran-comments.md`. This file holds details of our submission comments to CRAN and their responses to our submissions.  
 
 Change the `Version:` field in `DESCRIPTION` to the official version you want on CRAN (should not have a trailing `.9000`).
 
-This project uses Github Pages to host a documentation site:
+This project uses GitHub Pages to host a documentation site:
```
austin3dickey, line 156: TIL

.ci/setup.sh
```diff
@@ -11,7 +11,7 @@ export JAVA_APT_PKG="oracle-java8-set-default"
 # install these testing packages we need
 if [[ "$TASK" == "rpkg" ]];
 then
-  Rscript -e "install.packages(c('data.table', 'devtools', 'futile.logger', 'knitr', 'lintr', 'testthat', 'rmarkdown', 'uuid'), repos = 'http://cran.rstudio.com')"
+  Rscript -e "install.packages(c('data.table', 'futile.logger', 'knitr', 'lintr', 'testthat', 'rmarkdown', 'uuid'), repos = 'http://cran.rstudio.com')"
```
austin3dickey, line 14: We'll need to explicitly declare purrr here too since it apparently was being installed because of devtools. Can you make sure all of our [dependencies](https://github.com/uptake/uptasticsearch/blob/master/r-pkg/DESCRIPTION) are called out here?

jameslamb: ah! yep will do

jameslamb: updated! hopefully it works. Scary how many of these things we were getting just because of `devtools` :grimacing:

austin3dickey: Apparently we need the Suggests too, like covr... may be worth deleting covr from Suggests actually since we don't use it in CI.

austin3dickey: I guess we need Suggests too like covr!

jameslamb: I saw your comment in my email but I don't see it here...I agree we should just remove `covr` in `DESCRIPTION`. If we do that, we should probably do a minor release to CRAN after.

Remember the whole "`devtools::install()` installs all `Suggests` recursively" thing? :grimacing:

austin3dickey: Ha, sorry, I deleted it. I realized we *do* use covr in CI -- but we install it inside the script that calculates coverage. I was thinking it's better to leave it in Suggests and install it in CI.

jameslamb: oohhhhh ok got it. Yeah I'll make sure to add it. Guess it's another thing we  were getting from `devtools`

austin3dickey:
LGTM if the build passes once you fix conflicts!

CONTRIBUTING.md
```diff
@@ -191,7 +191,7 @@ Once the submission is accepted, great! Update `cran-comments.md` and merge the
 
 **Create a Release on GitHub**
 
-We use [the releases section](https://github.com/uptake/uptasticsearch/releases) in the repo to categorize certain important commits as release checkpoints. This makes it easier for developers to associate changes in the source code with the release history on CRAN, and enables features like `remotes::install_github()` for old versions.
+We use [the releases section](https://github.com/uptake/uptasticsearch/releases) in the repo to categorize certain important commits as release checkpoints. This makes it easier for developers to associate changes in the source code with the release history on CRAN, and enables features like `remotes:install_github()` for old versions.
```
jameslamb, line 194:
```suggestion
We use [the releases section](https://github.com/uptake/uptasticsearch/releases) in the repo to categorize certain important commits as release checkpoints. This makes it easier for developers to associate changes in the source code with the release history on CRAN, and enables features like `remotes::install_github()` for old versions.
```

PR #217, opened 2020-11-08
[ci] Move CI to GitHub Actions (fixes #215)

jameslamb:
This PR attempts to fix #215. See that issue and the links in it for details. Basically, Travis is becoming a lot harder for open source projects to use, and I think GitHub Actions is mature enough that we can move this project to it.

PR #220, opened 2021-05-24
[ci] update versions of ES, GitHub Actions

jameslamb:
It's been about 7 months since the last commit in this project (#217). This PR updates a few versions of things in CI as standard maintenance.

* Elasticsearch versions (based on https://www.elastic.co/downloads/past-releases)
* versions of all GitHub Actions used
* version of R (4.0.3 --> 4.1.0)

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -22,6 +22,8 @@ then
             qpdf \
         || exit -1
 
-    Rscript -e "install.packages(c('assertthat', 'covr', 'data.table', 'futile.logger', 'httr', 'jsonlite', 'knitr', 'lintr', 'purrr', 'rmarkdown', 'stringr', 'testthat', 'uuid'), repos = 'https://cran.r-project.org')"
+    conda install -y -c conda-forge 'pandoc>1.12.3'
+
+    Rscript -e "install.packages(c('assertthat', 'covr', 'data.table', 'futile.logger', 'httr', 'jsonlite', 'knitr', 'lintr', 'purrr', 'rmarkdown', 'stringr', 'testthat', 'uuid'), repos = 'https://cran.r-project.org'), Ncpus = parallel::detectCores())"
     cp test-data/* r-pkg/inst/testdata/
 fi
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -24,6 +24,6 @@ then
 
     conda install -y -c conda-forge 'pandoc>1.12.3'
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -69,19 +81,27 @@ jobs:
...
+          - task: pypkg
+            es_version: 7.10.2
+          - task: pypkg
+            es_version: 7.11.2
+          - task: pypkg
+            es_version: 7.12.1
     steps:
       - name: checkout repository
-        uses: actions/checkout@v2.3.4
+        uses: actions/checkout@v2
```
jameslamb, line 94: https://github.com/actions/checkout uses `v2` the way some projects use a `stable` tag. For example, if they publish a 2.3.5, at that moment `v2.3.5` and `v2` will point to the same commit.

So this is equivalent to doing something like `>=2.0.0,<3.0.0` in Python

austin3dickey (approved):

jameslamb:
R jobs are failing with the following

```text
Error: --- re-building ‘FAQ.Rmd’ using rmarkdown
Warning in engine$weave(file, quiet = quiet, encoding = enc) :
  Pandoc (>= 1.12.3) not available. Falling back to R Markdown v1.
Error: Error: processing vignette 'FAQ.Rmd' failed with diagnostics:
The 'markdown' package should be installed and declared as a dependency of the 'uptasticsearch' package (e.g., in the 'Suggests' field of DESCRIPTION), because the latter contains vignette(s) built with the 'markdown' package. Please see https://github.com/yihui/knitr/issues/1864 for more information.
--- failed re-building ‘FAQ.Rmd’

SUMMARY: processing the following file failed:
  ‘FAQ.Rmd’

Error: Error: Vignette re-building failed.
Execution halted
Error: Process completed with exit code 1.
```

According to https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md, `pandoc` isn't guaranteed to be available. I'll add it here.

.ci/setup.sh
```diff
@@ -22,6 +22,8 @@ then
             qpdf \
         || exit -1
 
-    Rscript -e "install.packages(c('assertthat', 'covr', 'data.table', 'futile.logger', 'httr', 'jsonlite', 'knitr', 'lintr', 'purrr', 'rmarkdown', 'stringr', 'testthat', 'uuid'), repos = 'https://cran.r-project.org')"
+    conda install -y -c conda-forge 'pandoc>1.12.3'
+
+    Rscript -e "install.packages(c('assertthat', 'covr', 'data.table', 'futile.logger', 'httr', 'jsonlite', 'knitr', 'lintr', 'purrr', 'rmarkdown', 'stringr', 'testthat', 'uuid'), repos = 'https://cran.r-project.org'), Ncpus = parallel::detectCores())"
```
jameslamb, line 27: installing packages is taking 8+ minutes (!!!), so I'm taking the opportunity to also add `Ncpus = parallel::detectCores()` here in `install.packages()`.

PR #224, opened 2025-01-11
remove Python package, drop Elasticsearch 1.0.3 support, get CI working again

jameslamb:
Removes the Python package.

Based on offline conversations with @austin3dickey and @ngparas , we've decided to retire the Python package here and focus this repo only on the R package (https://cran.r-project.org/web/packages/uptasticsearch/index.html).

The Python package was never distributed via package managers and I don't think it's in use anywhere today.

### Other changes to get CI working

**Updating syntax for `{lintr}`**

To fix stuff like this:

> *the use of linters of class 'function' was deprecated in lintr version 3.0.0. Use linters classed as 'linter' (see ?Linter) instead.*

**Skipping `{lintr}` checks that raise errors (I'll fix these in follow-up PRs).***

**Dropping Elasticsearch 1.0.3 support**

The Docker images we were using for that support can't run on newer versions of docker.

> Elasticsearch version: 1.0.3
Starting up Elasticsearch...
Unable to find image 'barnybug/elasticsearch:1.0.3' locally
1.0.3: Pulling from barnybug/elasticsearch
docker: [DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of docker.io/barnybug/elasticsearch:1.0.3 to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at [https://docs.docker.com/go/deprecated-image-specs/.](https://docs.docker.com/go/deprecated-image-specs/)
See 'docker run --help'.

([build link](https://github.com/uptake/uptasticsearch/actions/runs/12721809805/job/35465157559?pr=224#step:5:2585))

Seems ok... Elasticsearch 1.0.3 came out in April 2014 ([link](https://github.com/elastic/elasticsearch/releases/tag/v1.0.3)).


jameslamb:
yay CI works again 🎉 

All things considered... pretty cool that this is all it took after around 4 years of inactivity.

austin3dickey:
nice! 
image

PR #230, opened 2025-01-12
WIP: restructure CI

jameslamb:


jameslamb:
My motivation for this was "it seems wasteful to rebuild the package for every version of Elasticsearch we test against". I was gonna try having one build job that produced a package, and then testing jobs that all re-used it.

But honestly, I don't think it's worth the effort and complexity. most of the test-job time is dominated by installing *dependencies*, most of which overlap with the build-time dependencies either way.

Closing this.

PR #231, opened 2025-01-13
[ci] simplify CI

jameslamb:
Following #224, this project only produces an R package.

This proposes some simplifications to the CI setup based on no longer needing to have stuff like "if R do x".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,79 +10,57 @@ on:
 
 jobs:
   test:
-    name: ${{ matrix.task }} (ES ${{ matrix.es_version }})
+    name: test (ES ${{ matrix.es_version }})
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -10,79 +10,57 @@ on:
...
+    if: always()
     runs-on: ubuntu-latest
-    needs: [test]
+    needs:
+      - test
     steps:
-    - name: Note that all tests succeeded
-      run: echo "🎉"
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@v1.2.2
```
jameslamb, line 64: been using this for a few years across various projects, it works great:

https://github.com/orgs/community/discussions/26733#discussioncomment-4481488

PR #234, opened 2025-01-29
[ci] add more {lintr} linters

jameslamb:
follow-up to #233 

Adds more linters from `{lintr}`, to further standardize the codebase.

Most of the diff here is single quotes to double quotes, sorry 😅 

Fixes new linter findings like these:

```text
[class_equals] Instead of comparing class(x) with %in%, use inherits(x, 'class-name') or is.<class> or is(x, 'class')
[fixed_regex] This regular expression is static, i.e., its matches can be expressed as a fixed substring expression, which is faster to compute. Here, you can use "doc_count" with fixed = TRUE.
[line_length] Lines should not be more than 120 characters. This line is 126 characters.
[paste] Construct file paths with file.path(...) instead of paste(..., sep = "/")
[paste] toString(.) is more expressive than paste(., collapse = ", "). Note also glue::glue_collapse() and and::and() for constructing human-readable / translation-friendly lists
[quotes] Only use double-quotes.
[unnecessary_lambda] Pass is.na directly as a symbol to vapply() instead of wrapping it in an unnecessary anonymous function. For example, prefer lapply(DF, sum) to lapply(DF, function(x) sum(x)).
[vector_logic] Conditional expressions require scalar logical operators (&& and ||)
```


r-pkg/tests/testthat/test-unpack_nested_data.R
```diff
@@ -12,44 +12,48 @@ futile.logger::flog.threshold(0)
...
               expect_equivalent(dim(unpackedDT), c(7, 8))
-              expect_named(unpackedDT, c('dateTime', 'username', 'details.interactions',
-                                         'details.userType', 'appName', 'minutes', 'value', 'typovalue'))
-              expect_identical(unpackedDT$appName, c('farmville', 'candy_crush', 'angry_birds',
-                                                     'minesweeper', 'pokemon_go', 'pokemon_stay',
-                                                     'block_dude'))
+              expect_named(unpackedDT, c("dateTime", "username", "details.interactions",
+                                         "details.userType", "appName", "minutes", "value", "typovalue"))
+              expect_identical(unpackedDT$appName, c("farmville", "candy_crush", "angry_birds",
+                                                     "minesweeper", "pokemon_go", "pokemon_stay",
```
austin3dickey, line 33: omg

jameslamb: we were so funny, huh

Issue #229, opened 2025-01-12
Drop {httr} dependency


jameslamb:
## Description

Active development on `{httr}` stopped 2 years ago.

From its docs:

> *httr is superseded: only changes necessary to keep it on CRAN will be made. We recommend using [httr2](https://httr2.r-lib.org) instead.*

([r-lib / httr / README.md](https://github.com/r-lib/httr/blob/df11e216f5e1a30f1bc8df8118e7cd983d9066f9/README.md?plain=1#L7))

`{uptasticsearch}` should switch to `{httr2}` or something else.

## Benefits of this work

* reduces the risk of needing to update under thread of CRAN archival if / when `{httr}` is eventually fully abandoned
* allows `{uptasticsearch}` to take advantage of whatever improvements `{httr2}` has over `{httr}`

## Acceptance criteria

* {uptasticsearch} does not depend on `{httr}`

## Approach

We have good tests we can trust, try a change and see what happens.
Choose a replacement that's actively maintained and that hopefully does not introduce too many new dependencies.

## Notes

N/A

jameslamb:
Looked into `{httr2}`'s dependencies tonight.

```r
library(pkgnet)
install.packages("httr2")
pkgnet::CreatePackageReport("httr2")
```

It has 18 recursive strong (`Depends`, `Imports`, `LinkingTo`) dependencies

| package  | httr   | httr2  |
|:---------:|:-----:|:------|
| R6            |   ✔     | ✔      |
| askpass   |    ✔    | ✔      |
| cli             |            | ✔      |
| curl           |  ✔       | ✔      |
| glue          |            | ✔      |
| grDevices |  ✔      | ✔      |
| graphics   |   ✔     | ✔      |
| jsonlite     |   ✔      |       |
| lifecycle    |            | ✔      |
| magrittr     |            | ✔      |
| methods   |  ✔      | ✔      |
| mime         |  ✔      |          |
| openssl     |   ✔      | ✔      |
| rappdirs    |            | ✔      |
| rlang         |            | ✔      |
| stats         |    ✔     | ✔      |
| sys            |    ✔     | ✔      |
| tools          |    ✔     |       |
| utils           |    ✔      | ✔      |
| vctrs          |            | ✔      |
| withr          |            | ✔      |

So:

```text
+ cli, glue, lifecycle, magrittr, rappdirs, rlang, vctrs, withr
- jsonlite, mime, tools
```

[...]

Issue #228, opened 2025-01-12
[docs] switch vignettes from 'rmarkdown' to 'markdown'


jameslamb:
## Description

This project has one small vignette that doesn't rely on any special features of R markdown.

`{uptasticsearch}` should switch its rendering engine from `{rmarkdown}` to `{markdown}`, to reduce the number of dependencies that need to be installed to build the docs.

## Benefits of this work

* reduces the maintenance effort required to keep the package on CRAN
  - *e.g., reduces the risk of being affected by stuff like #223*
* reduces the burden that `{uptasticsearch}` places on other projects and on CRAN
  - *by removing `{uptasticsearch}` from more projects' reverse dependency lists*

## Acceptance criteria

* `{uptasticsearch}`'s docs can be built without `{rmarkdown}`

## Approach

See https://github.com/microsoft/LightGBM/pull/6258

## Notes

N/A

PR #237, opened 2025-01-29
remove uses of testthat::with_mock() (fixes #223)

### LINKED ISSUE #223: with_mock() is going away
This is a heads up to let you know that `with_mock()` is deprecated and will be made defunct soon. The API functions that made it work are being removed in R 4.5.0, so we'd suggest that you switch to `with_mocked_bindings()` instead.

jameslamb:
Fixes #223

This removes the one use of `testthat::with_mock()` in the package, for the reasons mentioned in #223.

It does this by moving the `{httr}` calls we want to mock into separate internal functions within `{uptasticsearch}`. Those thin wrappers are introduced here mainly because of the advice from https://testthat.r-lib.org/reference/local_mocked_bindings.html#namespaced-calls

> _It's trickier to mock functions in other packages that you call with `::`. For example, take this minor variation:_
>
> ```r
> some_function <- function() {
>   anotherpackage::another_function()
> }
> ```
>
> _To mock this function, you'd need to modify `another_function()` inside the `anotherpackage` package. You can do this by supplying the `.package` argument to `local_mocked_bindings()` but we don't recommend it because it will affect all calls to `anotherpackage::another_function()`, not just the calls originating in your package. Instead, it's safer to either import the function into your package, or make a wrapper that you can mock._

## Notes for Reviewers

These new internal functions are used in multiple files, so I wanted to centralize them in one other file. Introducing: `helperfuns.R` 😊

### DIFF
--- a/r-pkg/R/es_search.R
+++ b/r-pkg/R/es_search.R
@@ -222,7 +222,6 @@ es_search <- function(es_host
 # \item \href{http://stackoverflow.com/questions/25453872/why-does-this-elasticsearch-scan-and-scroll-keep-returning-the-same-scroll-id}{More background on how/why Elasticsearch generates and changes the scroll_id}
 # }
 #' @importFrom data.table rbindlist setkeyv
-#' @importFrom httr RETRY content
 #' @importFrom jsonlite fromJSON
 #' @importFrom parallel clusterMap makeForkCluster makePSOCKcluster stopCluster
 #' @importFrom uuid UUIDgenerate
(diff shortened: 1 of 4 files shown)


austin3dickey (approved):
makes sense to me!

jameslamb:
Thanks! This is the only change we **need* to be protected on CRAN ahead of R 4.5.0.

I expect that to come out in April ([as every previous minor release has](https://cran.r-project.org/bin/windows/base/old/)), so we have a bit more time. I'm going to continue with a few other things to try to reduce the we-get-a-fix-or-be-archived-email-from-CRAN risk.

Like these:

* #228
* #229

PR #240, opened 2025-01-30
drop {uuid} dependency (fixes #238)

### LINKED ISSUE #238: Drop {uuid} dependency
## Description

This project uses `{uuid}` in just 2 places, both of which could safely be dropped. Neither use requires the uniqueness guarantees of a UUID.

**use 1: temporary column name**

`{uuid}` is used to generate a temporary column name to make a join preserve ordering

https://github.com/uptake/uptasticsearch/blob/02b3d1751d30c4a46b41295586824eadce319768/r-pkg/R/unpack_nested_data.R#L60

which is later dropped

https://github.com/uptake/uptasticsearch/blob/02b3d1751d30c4a46b41295586824eadce319768/r-pkg/R/unpack_nested_data.R#L104

For that purpose, any column name not already in the data frame is sufficient... it doesn't have to be *universally unique*.

**use 2: file names**

Results are streamed out to local files as `{uptasticsearch}` pages over responses, to avoid out-of-memory errors.

https://github.com/uptake/uptasticsearch/blob/02b3d1751d30c4a46b41295586824eadce319768/r-pkg/R/es_search.R#L333

https://github.com/uptake/uptasticsearch/blob/02b3d1751d30c4a46b41295586824eadce319768/r-pkg/R/es_search.R#L387-L399

For that purpose, the files just need to not exist yet. base R's `tempdir()` and `tempfile()` would be totally sufficient for this.

## Benefits of this work

* removes a dependency (one less thing to compile on Linux, one less possible source of "your dependency was archived on CRAN so now you have to fix it or be archived to)

## Acceptance criteria

* `{uptasticsearch}` does not depend on `{uuid}` or any other third-party packages to generate random column names and file names

## Approach

Fact out `{uuid}`.

Remove `uuid` from `DESCRIPTION` and any scripts manually installing it.

## Notes

N/A

jameslamb:
Fixes #238

Removes this project's dependency on the `{uuid}` package.

As described in #238, I think we can confidently and performantly generate random strings that meet `{uptasticsearch}`'s needs without requiring a third-party package for generating UUIDs. The use-cases here just do not require the strong uniqueness guarantees that UUIDs provide.


jameslamb:
I'll leave this open for a few days to give you a chance to look, @austin3dickey . I'd like your opinion on some of these changes, if you have time and interest.

r-pkg/R/es_search.R
```diff
@@ -277,7 +276,8 @@ es_search <- function(es_host
 
     # Find a safe path to write to and create it
     repeat {
-        out_path <- file.path(intermediates_dir, uuid::UUIDgenerate())
+        random_dirname <- sprintf("tmp-%s", .random_string(36L))
+        out_path <- file.path(intermediates_dir, random_dirname)
         if (!dir.exists(out_path)) {
```
austin3dickey, line 281: Haha I was gonna write something about "what if someone had set a seed and not cleaned up old directories," since now `.random_string()` gives reproducible results if the seed was set, but I forgot how we protect against collisions with `repeat` and `break` 😂

austin3dickey (approved):
Looks great! Thanks for cleaning up deps!

r-pkg/R/es_search.R
jameslamb: Ha yes I DISTINCTLY remember when we talked about whether this was overkill when we first added it for paths 😂 

Excellent point about this being affected by random seed, I hadn't thought about that.

austin3dickey: Maybe it's worth a little comment in the helper func but other than that I think this is fine to ship!

jameslamb: sure! Added a little more description in https://github.com/uptake/uptasticsearch/pull/240/commits/8327487a3d803b0bf7224dd08aeac7382cd2db59. I'll merge this once CI passes.

PR #243, opened 2025-02-01
drop {assertthat} dependency

### LINKED ISSUE #242: Drop {assertthat} dependency
## Description

`{assertthat}` is a strong dependency of this `{uptasticsearch}`. It's a collection of small functions that provide some syntactic sugar and nice error messages for common assertions.

`{uptasticsearch}` should drop that dependency.

## Benefits of this work

* reduces the risk of needing to update under thread of CRAN archival if / when `{assertthat}` is eventually fully abandoned
* reduces the number of packages that need to be installed to use `{uptasticsearch}`
* removes a GPL-3 dependency from `{uptasticsearch}`, which could be a problem in some environments that are way of copyleft licenses

## Acceptance criteria

* {uptasticsearch} does not depend on `{assertthat}`

## Approach

We have good tests we can trust, try a change and see what happens.

Try to replace these with internal helper functions. 

## Notes

`{assertthat}` appears to be abandoned.

* last commit was almost 6 years ago: https://github.com/hadley/assertthat/commits/master/
* last release was almost 6 years ago: https://cran.r-project.org/web/packages/assertthat/index.html
* last time a maintainer responded to an issue was 4+ years ago: https://github.com/hadley/assertthat/issues

jameslamb:
Fixes #242

Proposes dropping the package's dependency on `{assertthat}`, for the reasons mentioned in #242


r-pkg/R/get_fields.R
```diff
@@ -27,9 +27,10 @@ get_fields <- function(es_host
     # Input checking
     es_url <- .ValidateAndFormatHost(es_host)
 
+    # other input checks with simple error messages
     .assert(
-        is.character("es_indices")
```
jameslamb, line 31: look look, this one was a bug!!!

It was always checking the literal string `"es_indices"` instead of checking the content of the argument `es_indices` 😅

austin3dickey (approved):
Nice!

PR #241, opened 2025-01-31
WIP: [docs] [ci] update website, move files out of source control

jameslamb:
Testing moving docs deployments to this pattern:

https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow

PR #244, opened 2025-02-08
[docs] [ci] update website, move docs files out of source control

jameslamb:
Testing moving docs deployments to this pattern:

https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow

Benefits:

* no need to check all these HTML/CSS/JS files into source control
* automatically keeps the docs up to date with the current state of the code (no need to manually re-generate it)

## Notes for Reviewers

Replaces #241.


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Tests
+name: ci
```
jameslamb, line 1: Now CI has more than just tests.

Also this shorter prefix makes it a little easier to read in the UI:

image

.github/workflows/ci.yml
```diff
@@ -18,6 +25,12 @@ env:
   MAKEFLAGS: "-j4"
 
 jobs:
+  build-docs:
+    uses: ./.github/workflows/build-docs.yaml
+    with:
+      #deploy: ${{ (github.event_name == 'push' && startsWith(github.ref, 'refs/tags')) || (github.event_name == 'workflow_dispatch' && inputs.deploy-docs == true) }}
+      deploy: true
```
jameslamb, line 32:
```suggestion
      deploy: ${{ (github.event_name == 'push' && startsWith(github.ref, 'refs/tags')) || (github.event_name == 'workflow_dispatch' && inputs.deploy-docs == true) }}
```

TODO: revert this before merging.

I'm proposing only updating the docs site on the builds triggered by merges to `main`. Just had this set to `true` to test this PR.

r-pkg/_pkgdown.yml
```diff
@@ -1,17 +1,25 @@
...
   url:
     home: https://github.com/uptake/uptasticsearch/
     source: https://github.com/uptake/uptasticsearch/tree/main/r-pkg/
     issue: https://github.com/uptake/uptasticsearch/issues
     user: https://github.com/
 
-
 reference:
-  - title: Main function
+  - title: Main function (DELETE ME)
```
jameslamb, line 22:
```suggestion
  - title: Main function
```

TODO: remove this before merging.

Put this here to prove that docs deployed this way are making it to the site: https://uptake.github.io/uptasticsearch/reference/index.html

image

.github/workflows/build-docs.yaml
```diff
@@ -0,0 +1,56 @@
...
+    inputs:
+      deploy:
+        required: true
+        type: boolean
+        default: false
+        description: "set to true to publish docs"
+
+jobs:
+  build:
+    runs-on: macos-latest
```
jameslamb, line 19: Proposing using macOS because CRAN has pre-built binaries there. So CI is faster and less likely to fail with could-not-build-some-dependency-from-source types of errors.

The docs build takes around 3 minutes to run on macOS: https://github.com/uptake/uptasticsearch/actions/runs/13212614464/job/36888065982?pr=244

austin3dickey (approved):
This is so much better! Thanks!!

jameslamb:
Thanks for taking a look!

PR #247, opened 2025-02-17
[ci] run linting in CI, add dependabot

jameslamb:
I realized today that this project's static analyzers (`pre-commit` hooks + `{lintr}` for R code) aren't actually being run in CI.

This fixes that.

* adds a CI job to run linting
* adds `dependabot` to automatically open PRs updating third-party GitHub Actions versions
* changes `lint_r_code.R` to `lint-r-code.R` (personal preference, fewer keypresses, it just makes me happy)
* fixes these `{lintr}` warnings (new as of `{lintr}` 3.2.0)

```text
[paste] Use paste(), not paste0(), to collapse a character vector when sep= is not used.
```

* fixes this `R CMD check` NOTE caused by #243 

```text
* checking R code for possible problems ... NOTE
.assert: warning in eval(expr, env = parent.frame()): partial argument
  match of 'env' to 'envir'
```

## Notes for Reviewers

dependabot will check for new versions of third-part GitHub Actions, and open PRs like this: https://github.com/jameslamb/pydistcheck/pull/298

This is useful on its own, but also for a repo like this that isn't being actively maintained... that'll give us an occasional CI run, which might reveal other stuff that needs to be updated to keep things minimally working here.


.github/workflows/ci.yml
```diff
@@ -79,6 +95,8 @@ jobs:
     if: always()
     runs-on: ubuntu-latest
     needs:
+      - build-docs
+      - lint
```
jameslamb, line 99: Should have added this in #244, so docs build failures will block PRs.

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+    schedule:
+      interval: monthly
+    groups:
+      ci-dependencies:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: "[ci]"
+    labels:
+      - maintenance
```
austin3dickey, line 15: This is awesome!

austin3dickey (approved):
looks great!

jameslamb:
Thanks for reviewing!

PR #249, opened 2025-02-23
replace {httr} with {curl}

### LINKED ISSUE #229: Drop {httr} dependency

jameslamb:
fixes #229 

* replaces `{httr}` with `{curl}`
* adds some integration tests covering HTTP-request code (that don't rely on Elasticsearch)

### How I tested this

The existing integration tests cover all the Elasticsearch interactions.

To check that the retry logic works, I built the package from source and installed it.

```shell
cd r-pkg/
R CMD INSTALL .
```

Then, in R:

```r
futile.logger::flog.threshold(futile.logger::DEBUG)

# request that should succeed
res <- uptasticsearch:::.request(
    verb = "GET"
    , url = "https://httpbin.org/status/429"
    , body = NULL
    , add_json_headers = FALSE
)
```

Saw logs like this:

```text
DEBUG [2025-02-22 23:07:40] Request failed (status code 429): 'GET https://httpbin.org/status/429'
DEBUG [2025-02-22 23:07:40] Sleeping for 1.88 before retrying.
DEBUG [2025-02-22 23:07:42] Request failed (status code 429): 'GET https://httpbin.org/status/429'
DEBUG [2025-02-22 23:07:42] Sleeping for 2.62 before retrying.
DEBUG [2025-02-22 23:07:45] Request failed (status code 429): 'GET https://httpbin.org/status/429'
```

And a response object returned:

```r
res$method
# [1] "GET"

res$status_code
# [1] 429

res$url
# [1] "https://httpbin.org/status/429"
```


r-pkg/R/get_fields.R
```diff
@@ -243,52 +242,13 @@ get_fields <- function(es_host
...
         return(invisible(NULL))
     } else {
-        major_version <- .get_es_version(es_host)
-        process_alias <- switch(
-            major_version
-            , "1" = .process_legacy_alias
-            , "2" = .process_legacy_alias
-            , "5" = .process_new_alias
-            , "6" = .process_new_alias
-            , .process_new_alias
```
jameslamb, line 253: Turns out we didn't need any of this stuff!!!

I realized through this exercise that the difference across Elasticsearch versions was like this:

* we weren't passing the `Accept` header in this request to `GET /_cat/aliases`
* on ES 1.x and 2.x, `GET /_cat/aliases` responded with fixed-width text by default
* from ES 3.x onwards, it responded with JSON data by default
* across ALL versions, it responds with JSON data if you pass `Accept: application/json`

So just passing that header, this now always returns JSON and no more need for these different processing functions 😁 😁 😁

r-pkg/R/helperfuns.R
```diff
@@ -21,28 +32,135 @@
...
+        if (attempt_count > 1L) {
+            # exponential backoff with jitter
+            #
+            #   1.45s + {jitter}
+            #   2.10s + {jitter}
+            #   3.05s + {jitter}
+            #   etc., etc.
+            #
+            # ref: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+            sleep_seconds <- 1.45 ** (attempt_count - 1L) + stats::runif(n = 1L, min = 0.1, max = 0.5)
```
jameslamb, line 84: 1.45s and the jitter being 0.1-0.5 are arbitrary. @austin3dickey if you have a different suggestion for these I'd take them, I really don't feel that strongly about it.

r-pkg/R/get_fields.R
austin3dickey: Amazing!

r-pkg/R/helperfuns.R
austin3dickey: nah this seems in line with what's normally done!

r-pkg/R/helperfuns.R
```diff
@@ -21,28 +32,135 @@
...
-    result <- httr::RETRY(
-        verb = verb
+#' @importFrom curl handle_setheaders handle_setopt new_handle
+.request <- function(verb, url, body) {
+    handle <- curl::new_handle()
+
+    # set headers
+    #
+    # This can safely be hard-coded here because every payload this library
+    # posts and every response body it receives is JSON data.
```
austin3dickey, line 123: Cool!

r-pkg/tests/testthat/test-integration.R
```diff
@@ -474,5 +473,42 @@ ES_HOST <- "http://127.0.0.1:9200"
...
+})
+
+test_that("retry logic works as expected", {
+    testthat::skip_on_cran()
+    futile.logger::flog.threshold(futile.logger::DEBUG)
+    log_lines <- testthat::capture_output({
+        response <- .request(
+            verb = "GET"
+            , url = "https://httpbin.org/status/502"
+            , body = NULL
```
austin3dickey, line 496: I forget, does one of the other integration tests use a body? If not maybe we should add one just for a lil more coverage

austin3dickey (approved):
Looks great!

r-pkg/tests/testthat/test-integration.R
jameslamb: Many do use it implicitly... for example `es_search()` is going to issue a `POST` request with the query attached as a body.

But agree it'd be good to have an explicit one here, to cover that `if()` block. I'll add that before merging.

r-pkg/R/helperfuns.R
jameslamb: ok great, thank you

r-pkg/tests/testthat/test-integration.R
jameslamb: Just pushed https://github.com/uptake/uptasticsearch/pull/249/commits/51ee836553ce88f6efb2cebe296fc619c6db8aed with 2 more tests, explicitly covering the case where requests have a body.

jameslamb:
Those new tests are passing, `R CMD check` looks happy... onwards, to CRAN!!

Thanks as always @austin3dickey

PR #254, opened 2025-02-24
[docs] remove defunct emails from DESCRIPTION

jameslamb:
A few authors still have `@uptake.com` emails in `DESCRIPTION`. None of those people work there any more, so those emails can't be used to reach those people.

This proposes just removing the email addresses and not replacing them. CRAN only requires an email address for the maintainer.

### DIFF
--- a/r-pkg/DESCRIPTION
+++ b/r-pkg/DESCRIPTION
@@ -4,11 +4,11 @@ Title: Get Data Frame Representations of 'Elasticsearch' Results
 Version: 0.4.0.9999
 Authors@R: c(
     person("James", "Lamb", email = "jaylamb20@gmail.com", role = c("aut", "cre")),
-    person("Nick", "Paras", email = "nick.paras@uptake.com", role = c("aut")),
-    person("Austin", "Dickey", email = "austin.dickey@uptake.com", role = c("aut")),
+    person("Nick", "Paras", role = c("aut")),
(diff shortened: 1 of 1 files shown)


r-pkg/DESCRIPTION
```diff
@@ -4,11 +4,11 @@ Title: Get Data Frame Representations of 'Elasticsearch' Results
 Version: 0.4.0.9999
 Authors@R: c(
     person("James", "Lamb", email = "jaylamb20@gmail.com", role = c("aut", "cre")),
-    person("Nick", "Paras", email = "nick.paras@uptake.com", role = c("aut")),
-    person("Austin", "Dickey", email = "austin.dickey@uptake.com", role = c("aut")),
+    person("Nick", "Paras", role = c("aut")),
+    person("Austin", "Dickey", role = c("aut")),
```
jameslamb, line 8: @austin3dickey let me know if you want a different email added here! I just assumed you wouldn't, but happy to add one.

austin3dickey: Yeah, probably not. I still follow Issues closely if people want to reach out 🙂

austin3dickey (approved):
nice change

r-pkg/DESCRIPTION
jameslamb: alright I'll leave it like this, thanks

PR #256, opened 2025-07-07
[ci] update pre-commit hooks

jameslamb:
Most just poking this to make sure CI is still working ok.

* updates `pre-commit` hooks with `pre-commit autoupdate`
* removes some commented-out configuration


austin3dickey (approved):
🪇

